PR #11, opened 2015-06-11
Update README.md

keokilee:
Fixed a typo.

I'm actually not sure about the wording for this line. Should it be "For tutorials and plugins"?


yasserf:
Many thanks for the typo fix, it has been fixed internally so won't be able to merge. Your comment on the wording is totally correct, have updated it accordingly!

PR #17, opened 2015-06-15
=BG= spelling

bguiz:


yasserf:
Many thanks for the correction, it has been fixed internally so won't be able to merge.

PR #33, opened 2015-08-10
SSL CA option

jwanglof:
We have an SSL certificate from GoDaddy where we need to provide an CA (Certificate authority) file. I tried to only add a key and certificate as sslKey and sslCert but this didn't work. With these patches added our SSL works fine. Don't know how to create any test(s) for this case, I'm afraid.

### DIFF
--- a/src/default-options.js
+++ b/src/default-options.js
@@ -25,6 +25,7 @@ exports.get = function() {
 		 */
 		sslKey: null,
 		sslCert: null,
+		sslCa: null,
 
 		/*
 		 * Data Manipulation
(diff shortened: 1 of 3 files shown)


yasserf:
Our tests around https are to ensure that the correct parameters are passed to createServer.

In [connection-endpointSpec.js](https://github.com/hoxton-one/deepstream.io/blob/master/test/message/connection-endpointSpec.js#L50-56) we check that both sslKey and sslCert has been provided as parameters to createServer .

If you could add a separate test case to indicate that if sslCa is provided it is also passed in as an argument we can then merge this.

Many thanks for the PR!

jwanglof:
Added three different tests for the new sslCa-option. Hope these are enough :)
Thanks for an awesome library!

yasserf:
Thanks for the quick update, Looks good!

PR #41, opened 2015-09-27
Event listening

yasserf:


yasserf:
@WolframHempel

jwanglof:
Isn't this already implemented in `client.event.subscribe( 'news/politics', function( news ), callback)`?

yasserf:
@jwanglof This is to mimic the record.listen functionality.

It's mainly for 'providers' ( any client that publishes data ) to allow them to only emit data only when at least one client is interested.

so its:
```javascript
client.event.listen( 'news/*', function( subject, subscribed ) {
    if( subscribed ) {
       // start emitting relevant events;
    } else {
       // stop emitting relevant events;
   }
} )
```

jwanglof:
Ah, very nice!

PR #47, opened 2015-11-26
Adding ability to exclude records from storage

### LINKED ISSUE #46: Select records to automatically store within database
Given that some records can be updated multiple times a second, this can result in putting the database under a bit of stress from continuous writing.

As such it would be ideal to have a filter system where we can optionally specify what records to store and which ones to ignore.

This can be done on a per record basis via the metadata associated when created, or potentially via a regex.

The following is just for reference, and not the decided API.

For example globals ( not as  fast, but alot easier and resistant to errors ):
```javascript
ds.set( 'storage-options', {
 include:  [ /store-me\//, /store-those\// ], //Only save if matches
 exclude:  [ /disposable\//, /too-quick\// ]//Only ignore if matches
} );
```

and per record, which is quicker ( and the API for meta data is still in the works ): 

```javascript
ds.record.getRecord( 'a-record', {
   store: false
} );
```

yasserf:
Adding ability to exclude records from storage to reduce unrequired database write -- Fixes #46

### DIFF
--- a/src/record/record-handler.js
+++ b/src/record/record-handler.js
@@ -154,12 +154,14 @@ RecordHandler.prototype._create = function( recordName, socketWrapper ) {
 		}
 	}.bind( this ));
 
-	// store the record data in the persistant storage independently and don't wait for the result
-	this._options.storage.set( recordName, record, function( error ) {
-		if( error ) {
-			this._options.logger.log( C.TOPIC.RECORD,  C.EVENT.RECORD_CREATE_ERROR, 'storage:' + error );
-		}
-	}.bind( this ) );
+	if( !this._options.storageExclusion || !this._options.storageExclusion.test( recordName ) ) {
+		// store the record data in the persistant storage independently and don't wait for the result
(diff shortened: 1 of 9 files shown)


src/record/record-handler.js
```diff
@@ -154,12 +154,14 @@ RecordHandler.prototype._create = function( recordName, socketWrapper ) {
 		}
 	}.bind( this ));
 
-	// store the record data in the persistant storage independently and don't wait for the result
-	this._options.storage.set( recordName, record, function( error ) {
-		if( error ) {
-			this._options.logger.log( C.TOPIC.RECORD,  C.EVENT.RECORD_CREATE_ERROR, 'storage:' + error );
-		}
-	}.bind( this ) );
+	if( !this._options.storageExclusion || !this._options.storageExclusion.test( recordName ) ) {
```
WolframHempel: Very minor (almost feel embarrassed), but personally I would have just gone with
```javascript
if( ... ) {
    return;
}
```
here to avoid the extra level of nesting

PR #89, opened 2016-02-12
Support for data in login response

### LINKED ISSUE #88: Allow returning data on valid auth
We should allow returning data on valid login attempts as well, not just invalid login attempts. E.g a JWT token:
```javascript
server.set('permissionHandler', {
  isValidUser: function(connectionData, authData, callback) {
    // authData contains { username, password }
    var token = jwt.authenticate(authData) // Returns a token

    callback(null, authData.username, token) // Sends the token back to the user
  }
})
```

Another example can be to return the users' profile data or.

Svenskunganka:
This adds support for data in login response, e.g:
```javascript
server.set('permissionHandler', {
  isValidUser: function(connectionData, authData, callback) {
    callback(null, authData.username, { property: 'value' })
  }
})
```

Fixes #88

### DIFF
--- a/src/message/connection-endpoint.js
+++ b/src/message/connection-endpoint.js
@@ -17,9 +17,9 @@ var C = require( '../constants/constants' ),
  * forwards messages it receives from authenticated sockets.
  *
  * @constructor
- * 
+ *
  * @extends events.EventEmitter
- * 
+ *
  * @param {Object} options the extended default options
  * @param {Function} readyCallback will be invoked once both the engineIo and the tcp connection are established
  */
@@ -73,7 +73,7 @@ util.inherits( ConnectionEndpoint, events.EventEmitter );
  * @param   {String} message the raw message as sent by the client
  *
  * @public
- * 
+ *
  * @returns {void}
  */
 ConnectionEndpoint.prototype.onMessage = function( socketWrapper, message ) {};
@@ -81,7 +81,7 @@ ConnectionEndpoint.prototype.onMessage = function( socketWrapper, message ) {};
 /**
  * Closes both the engine.io connection and the tcp connection. The ConnectionEndpoint
  * will emit a close event once both are succesfully shut down
- * 
+ *
  * @public
  * @returns {void}
  */
@@ -98,7 +98,7 @@ ConnectionEndpoint.prototype.close = function() {
 
 	this._engineIo.close();
 	this._server.close(function(){ this._engineIoServerClosed = true; }.bind( this ));
-	
+
 	// Close the tcp server
 	this._tcpEndpoint.on( 'close', this._checkClosed.bind( this ) );
 	this._tcpEndpoint.close();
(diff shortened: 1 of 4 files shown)


src/message/message-builder.js
```diff
@@ -2,7 +2,7 @@ var C = require( '../constants/constants' ),
 	SEP = C.MESSAGE_PART_SEPERATOR;
```
yasserf: This file doesn't have any changes from what I can see, might be good to exclude it from the commit

src/message/connection-endpoint.js
```diff
@@ -293,14 +294,15 @@ ConnectionEndpoint.prototype._processInvalidAuth = function( authError, authData
...
  * @private
- * 
+ *
  * @returns {void}
  */
-ConnectionEndpoint.prototype._processAuthResult = function( authData, socketWrapper, authError, username ) {
+ConnectionEndpoint.prototype._processAuthResult = function( authData, socketWrapper, authError, username, data ) {
 	if( authError === null ) {
-		this._registerAuthenticatedSocket( socketWrapper, username );
+		this._registerAuthenticatedSocket( socketWrapper, username, ( Array.isArray( data ) ) ? data : [ data ] );
```
yasserf: I think this may be incorrect. The following is the algo for creating messages. You always want your arguments to be a single part of the message.

So if you run:

`getMsg( 'A', 'A', [ 1, 2, 3 ] );`

you'll get:

> A|A|1|2|3+

instead of;

> A|A|[1,2,3]+

So you should probably be able to run it just using:

`this._registerAuthenticatedSocket( socketWrapper, username, [ data ] );`

getMsgAlgo:
```
exports.getMsg = function( topic, action, data ) {
	var sendData = [ topic, action ],
		i;
	if( data ) {
		for( i = 0; i < data.length; i++ ) {
			if( typeof data[ i ] === 'object' ) {
				sendData.push( JSON.stringify( data[ i ] ) );
			} else {
				sendData.push( data[ i ] );
			}
		}
	}
	return sendData.join( SEP ) + C.MESSAGE_SEPERATOR;
};
```

test/message/connection-endpointSpec.js
```diff
@@ -227,7 +227,7 @@ describe( 'connection endpoint', function() {
 
 			socketMock.emit( 'message', _msg( 'A|REQ|{"user":"wolfram"}+' ) );
 
-			expect( socketMock.lastSendMessage ).toBe( _msg( 'A|A+' ) );
```
yasserf: Would be good to keep the previous test and create a new one that returns both the user and data to ensure both scenarios work.

In the PermissionHandlerMock you can add a condition to return the params only when needed. 

To stick with the approach used, you would add a 

`permissionHandlerMock.nextUserValidationReturnData = 'test-data'`

and that would allow you to return the data you assigned, or undefined by default.

```
if( this.nextUserValidationResult === true ) {
    callback( null, 'test-user', nextUserValidationReturnData  );
 }
```

Issue #87, opened 2016-02-11
Attach to existing server (and, optionally, path)


inxilpro:
See https://github.com/deepstreamIO/deepstream.io-client-js/issues/63

Deepstream should be able to attach to an existing HTTP server and optionally attach to a non-root path.

yasserf:
This seems to be a duplicated of #95 which has been closed

PR #92, opened 2016-02-14
Allow for passing in an existing HTTP server

inxilpro:
Implements https://github.com/deepstreamIO/deepstream.io/issues/87 and the server component of https://github.com/deepstreamIO/deepstream.io-client-js/issues/63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ node_modules
 
 # Users Environment Variables
 .lock-wscript
+
+# IDE configuration
+.idea
--- a/src/default-options.js
+++ b/src/default-options.js
@@ -19,6 +19,8 @@ exports.get = function() {
 		host: argv.host || '0.0.0.0',
 		tcpPort: argv.tcpPort || 6021,
 		tcpHost: argv.tcpHost || '0.0.0.0',
+		httpServer: null,
+		urlPath: '/engine.io',
 
 		/*
 		 * SSL Configuration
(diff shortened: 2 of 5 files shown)


yasserf:
Nice!

Could you add tests in:

https://github.com/deepstreamIO/deepstream.io/blob/master/test/message/connection-endpointSpec.js

So that we know the passed in server has listen called on it? We already have a mockHttpServer so it should make things alot easier.

 https://github.com/deepstreamIO/deepstream.io/blob/master/test/message/connection-endpointSpec.js#L4

yasserf:
@WolframHempe @inxilpro l Should we have config sanity checks? There are a few option combinations now for creating a server.. if someone passes in the ssl cert and a httpserver it means something has been configured wrong

yasserf:
deepstreamIO/deepstream.io-client-js#63 would still require the functionality to no longer provide a url when being created and instead piggy back the pages root url + the url path configured

inxilpro:
@yasserf Just added some tests. The connection endpoint doesn't directly call `listen()`, but I added tests to make sure `_checkReady()` gets called regardless of whether `listen()` is called *before* or *after* the server is passed to deepstream.

That means both of these should work:

```javascript
// Listen second
var server = http.createServer();
deepstream.set('httpServer', server);
server.listen(6020);

// Listen first
var server = http.createServer();
server.listen(6020);
deepstream.set('httpServer', server);
```

yasserf:
Awesome thanks!

We just fixed #94 as well, so there are going to be a couple of merge conflicts unfortunately, but we'll merge your stuff in and sort that out as part of the merge.

CLClark:
this was deprecated... right?

yasserf:
Yup, was removed in V2

Issue #110, opened 2016-03-26
Add record snapshot support


yasserf:
We want to be able to request a snapshot from deepstream which returns the current state of the object in the cache or database without subscribing to following changes. 

This will be useful for simple single requests ( such as a HTTP endpoint with a GET ) and one off operations that don't require realtime updates.

In the case of clients with cache ( such as the js client ) if the record has already been subscribed to this will return whatever is in memory to save the server round trip.

# API
```
ds.record.snapshot( 'recordName', function( data ) {
    if( data === null ) {
        //record doesn't exist
        //return;
    }
    // do something with data
} );
```

# Requirements for completion

1) Server code 
2) Client code ( JS lib ) 
3) Update message spec 
4) Update client documentation

yasserf:
This is missing requirement 3, opening related issue in message spec

Issue #84, opened 2016-02-08
Add support for ds.record.has


WolframHempel:
This story spans both server and client

Let's add support for 
```javascript
ds.record.has( <string> name, function( <bool> hasRecord ){

});
```
to check if a certain record exists in cache or storage. **Important** this doesn't take the ability of dataproviders to provide the record into account.

WolframHempel:
Added in 0.9.0

PR #112, opened 2016-03-27
#84 and #110

yasserf:


src/record/record-handler.js
```diff
@@ -106,6 +120,47 @@ RecordHandler.prototype.handle = function( socketWrapper, message ) {
...
+ * @param   {SocketWrapper} socketWrapper the socket that send the request
+ * @param   {Object} message parsed and validated message
+ *
+ * @private
+ * @returns {void}
+ */
+RecordHandler.prototype._hasRecord = function( socketWrapper, message ) {
+	var recordName = message.data[ 0 ],
+		onComplete = function( record ) {
+			var hasRecord = record ? 'T' : 'F';
```
WolframHempel: Could we use `C.TYPES.TRUE` and C.`TYPES.FALSE`

yasserf: Yup!

src/record/record-handler.js
```diff
@@ -106,6 +120,47 @@ RecordHandler.prototype.handle = function( socketWrapper, message ) {
...
+			socketWrapper.sendMessage( C.TOPIC.RECORD, C.ACTIONS.HAS, [ recordName, hasRecord ] );
+		};
+
+	new RecordRequest( recordName, this._options, socketWrapper, onComplete.bind( this ) );
```
WolframHempel: Do we rely on the request timing out if an error occurred? Maybe rather pass an error callback.

src/record/record-handler.js
```diff
@@ -106,6 +120,47 @@ RecordHandler.prototype.handle = function( socketWrapper, message ) {
...
+ * @private
+ * @returns {void}
+ */
+RecordHandler.prototype._snapshot = function( socketWrapper, message ) {
+	var recordName = message.data[ 0 ],
+		onComplete = function( record ) {
+			if( record ) {
+				this._sendRecord( recordName, record, socketWrapper );
+			} else {
+				//TODO: This needs to be improved
```
WolframHempel: Agreed. Maybe we should rather send an error than introducing a new concept of null for missing

src/record/record-handler.js
```diff
@@ -176,6 +231,20 @@ RecordHandler.prototype._create = function( recordName, socketWrapper ) {
  */
 RecordHandler.prototype._read = function( recordName, record, socketWrapper ) {
 	this._subscriptionRegistry.subscribe( recordName, socketWrapper );
+	this._sendRecord( recordName, record, socketWrapper );
+};
+
+/**
+ * Sends the records data current data once done
```
WolframHempel: Hm - what changed here?

src/record/record-handler.js
yasserf: Is it an error though? Its correct to request something that doesn't exist

test/record/record-handlerSpec.js
```diff
@@ -62,6 +62,51 @@ describe( 'record handler handles messages', function(){
 		expect( clientA.socket.lastSendMessage ).toBe( msg( 'R|R|existingRecord|3|{"firstname":"Wolfram"}+' ));
 	});
 
+	it( 'returns true for HAS if message exists', function(){
+		recordHandler.handle( clientA, {
+			raw: msg( 'R|H|existingRecord' ),
+			topic: 'RECORD',
```
WolframHempel: topic: 'R'

src/record/record-handler.js
yasserf: Read now is a combination of subscribe and snapshot

test/record/record-handlerSpec.js
```diff
@@ -62,6 +62,51 @@ describe( 'record handler handles messages', function(){
...
+		recordHandler.handle( clientA, {
+			raw: msg( 'R|H|nonExistingRecord' ),
+			topic: 'RECORD',
+			action: 'H',
+			data: [ 'nonExistingRecord' ]
+		});
+
+		expect( clientA.socket.lastSendMessage ).toBe( msg( 'R|H|nonExistingRecord|F+' ) );
+	});
+
```
WolframHempel: Could we add a test for a failing record request

test/record/record-handlerSpec.js
yasserf: Good catch, I think I copied that from somewhere so will need to take a look where

src/record/record-handler.js
yasserf: Done

test/record/record-handlerSpec.js
yasserf: Done

PR #115, opened 2016-04-01
Bug/max message size

yasserf:
To be merged after https://github.com/deepstreamIO/deepstream.io/pull/112


src/tcp/tcp-socket.js
```diff
@@ -106,6 +113,33 @@ TcpSocket.prototype._onData = function( packet ) {
...
+ * 
+ *
+ * @emits 	{String} message
+ *
+ * @private
+ * @returns {void}
+ */
+TcpSocket.prototype._maxMessageSizeExceed = function() {
+	var errorMsg, indexEOM;
+
```
WolframHempel: I feel that this should be on the connection layer as both engine.io and TCP are affected

PR #116, opened 2016-04-02
Feature/listen snapshot

yasserf:
Given a snapshot would be used in a one off request ( ideally via HTTP and such going forward ) it makes sense to return a list of the array names instead of the individual ones.

I wouldn't merge this merge request until we release the others since the JS client work is outstanding and questionable in whether it needs to be implemented.


src/record/record-handler.js
```diff
@@ -106,6 +127,53 @@ RecordHandler.prototype.handle = function( socketWrapper, message ) {
...
+	var recordName = message.data[ 0 ],
+		onComplete = function( record ) {
+			var hasRecord = record ? C.TYPES.TRUE : C.TYPES.FALSE;
+			socketWrapper.sendMessage( C.TOPIC.RECORD, C.ACTIONS.HAS, [ recordName, hasRecord ] );
+		},
+		onError = function( error ) {
+			socketWrapper.sendError( C.TOPIC.RECORD, C.ACTIONS.HAS, [ recordName, error ] );
+		}
+
+	new RecordRequest( recordName, this._options, socketWrapper, onComplete.bind( this ), onError.bind( this ) );
```
WolframHempel: `onComplete()` and `onError()` shouldn't need the `bind( this )`

PR #137, opened 2016-05-06
Adding record version and data to VERSION_EXISTS error #1

yasserf:


src/record/record-transition.js
```diff
@@ -67,6 +68,31 @@ RecordTransition.prototype.hasVersion = function( version ) {
...
+ * Send version exists error if the record has been already loaded, else
+ * store the version exists error to send to the sockerWrapper once the 
+ * record is loaded
+ *
+ * @param   {SocketWrapper} socketWrapper the sender
+ * @param   {Number} version The version number
+ *
+ * @public
+ */
+RecordTransition.prototype.sendVersionExists = function( socketWrapper, version ) {
```
WolframHempel: I'm a bit confused: How do we know the record's version before it has been loaded?

yasserf: This was based on the previous logic, which is that the last version in the steps ( updates added  ) is assumed to be the records current version. I'm guessing this was to assume a positive path ( which is usually correct ).

test/record/record-handlerSpec.js
```diff
@@ -300,29 +303,68 @@ describe( 'record handler handles messages', function(){
...
 		recordHandler.handle( clientB, {
 			raw: msg( 'R|U|existingRecord|7|{"name":"Kowalski"}' ),
 			topic: 'R',
 			action: 'U',
 			data: [ 'existingRecord', 7, '{"name":"Kowalski"}' ]
 		});
 
+		setTimeout(function(){
+			expect( clientA.socket.lastSendMessage ).toBeNull();
+			/**
```
yasserf: @WolframHempel Be good to look at this behaviour / tests

test/record/record-requestSpec.js
```diff
@@ -75,11 +75,12 @@ describe( 'records are requested from cache and storage sequentually', function(
 		expect( options.storage.lastRequestedKey ).toBe( 'doesNotExist' );
 	});
 
-	it( 'fails gracefully if an error occured out of order', function( done ){
+	//TODO: yasserf: I don't get this test
```
WolframHempel: it destroys the record request in between getting data from the cache and from the storage - which is not supposed to happen...but just in case

yasserf: I meant more that done is called in the record request and and after 20 milliseconds, which means it can exit successfully without any expectations run

test/record/record-handlerSpec.js
WolframHempel: on it

PR #139, opened 2016-05-07
Code Coverage at 100%

WolframHempel:
Sorry for amount of changes, but I've also removed whitespace, unnecessary functions, added semicolons and other bits I found along the way.

### DIFF
--- a/test/utils/dependency-initialiserSpec.js
+++ b/test/utils/dependency-initialiserSpec.js
@@ -12,9 +12,8 @@ describe( 'dependency-initialiser', function(){
 		pluginB: new PluginMock( 'B' ),
 		pluginC: new PluginMock( 'C' ),
 		logger: new LoggerMock(),
-		dependencyInitialisationTimeout: 20
+		dependencyInitialisationTimeout: 10
 	};
-
 
 	it( 'selects the correct plugin', function(){
 		dependencyBInitialiser = new DependencyInitialiser( options, 'pluginB' );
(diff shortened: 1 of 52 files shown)


test/utils/dependency-initialiserSpec.js
```diff
@@ -27,11 +26,74 @@ describe( 'dependency-initialiser', function(){
...
+	var log = jasmine.createSpy( 'log' );
+	var originalProcessExit = process.exit;
+	var originalConsoleLog = console.log;
+	var options = {
+		plugin: new PluginMock( 'A' ),
+		logger: { log: jasmine.createSpy( 'log' ), isReady: true },
+		dependencyInitialisationTimeout: 1
+	};
+
+	it( 'disables process exit', function(){
```
yasserf: these should be beforeAll and afterAll since they don't actually validate anything. Just saying =D

WolframHempel: They used to be - it breaks (try it if you like). Something is different in their timing, it is called before process.exit is called,

yasserf: Interesting.

WolframHempel:
@adityamenon Good spot, but our problem actually is that we run our build with multiple versions of node, all the way back to 0.12 - which doesn't support ES6 syntax yet.

At some point we will abandon support for Node versions <1, but I feel it needs to be for a more crucial reason than me being to lazy to type `function` :-)

PR #147, opened 2016-05-12
Feature/permission config

### LINKED ISSUE #125: Configuration based Permissions
Why do we need this?
===================================
[deepstream's current, functional style of permissioning](https://deepstream.io/tutorials/permissioning.html) was build to make it easy to integrate with existing user databases, active directories or other rights management systems - but it has proven to be hard to use for many developers and requires quite a bit of code if you don't want to use an external permission system. Configuration based permissions make it easier to create read, write and validation rules using a simple JSON structure.

Will this replace the current permissioning system?
===========================================
No. Programmatic permissions still fulfill an important function for integrating with third party systems. Both approaches will live side by side and its up to the developer to choose the one best suited for their usecase.

What will the permission system look like?
===========================================
deepstream's permissions will be inspired by [Firebase's Bolt](https://github.com/firebase/bolt) approach (Apache 2 Licensed). There will however be fundamental differences (as there are fundamental differences between deepstream and firebase) and they won't be directly compatible.

Here's what a deepstream `permissions.json` file might look like
```javascript
{
    "record": {
        "*": {
            "all": true,
            "listen": false
        },
        "game-score/$gameId": {
            //publicly readable
            "read": true,
            //writable only by provider
            "write": "user.data.role === 'provider'"
        },

        "game-comment/$gameId/$userId/$commentId": {
            // readable by all logged in (authenticated) users
            "read": "user.isAuthenticated === true",
            // Users can only create / edit comments with their own userid
            // Each comment must have a title and a content field that are strings
            "write": "user.id === $userId && typeof data.title ===  'string' && data.content ===  'string"
        }
    },
    "event": {
        "*": {
            "publish": false,
            "subscribe": false
        }
    },
    "rpc": {
        "*": {
            "provide": false,
            "request": false
        }
    }
}
```

What can be validated?
====================================
Records, Events and RPCs. Records will be treated as atomic units, sub-paths within records can be validated via the write rule, e.g.

```json
{
    "write": "data.car.price > 5000"
}
```

Which rule types are supported?
==================================
For records, you can permission `create`, `read`, `write`, `delete` and `listen`, for events `publish`,  `subscribe` and `listen` and for rpcs `request` and `response`

Only `record:write`, `event.publish` and `rpc.request` support data. 

Only `record.write`, `record.read` and `record.delete` support oldData

How do rules work?
==================================
Rules take the form of either true (always allowed), false (never allowed) or a javascript string that evaluates to true or falls. String rules don't support functions with the exception of `_()` for cross-referencing and the string functions `startsWith, endsWith, includes, indexOf, match, toUpperCase, toLowerCase, trim`. A number of variables are accessible within the rule, more about that below.

Examples of valid rules:
```javascript
true
false
'1 === 1'
'user.id === 14'
'user.id === $userId'
'user.data.role === "admin"'
'data.lastname.toLowerCase() === "doe"'
'_("item/" + data.itemId).price < data.price'
```

Examples of invalid rules
```javascript
14
'a=new SomeClass'
'data.lastname.toUpperCase() && data.lastname.substr(0,3)'
'alert("bobo")'
'a b'
```

Which variables are supported?
==================================
#### user
the authentication data for the user attempting the read or write

```javascript
{
    //Boolean, false for unauthenticated users
    isAuthenticated: <Boolean>,
    //the userid as returned by auth the auth provider
    id: <String>,
    //optional object, containing fields like e.g. role, access level etc
    //returned by auth provider
    data: <Object>
}
```

#### data
the incoming data for records, events and rpcs

#### oldData
the current data, only for records

#### now
current timestamp on the server

#### action
the original action that triggered this rule (e.g. UPDATE / PATCH / LISTEN ) etc. Useful for more finegrained/low-level permissions

#### $&lt;variableName&gt;
Variables that are extruded from the record / event / rpc name. Names can contain multiple variables. Variable names start with a dollar and are only allowed to contain uppercase letters, lowercase letters and numbers.

```javascript
{
    "record": {
        //make sure users can only manipulate their own profile
        "user-profile/$userId": {
            "write": "$userId === user.id"
        }
    }
}
```

#### _(&lt;recordName&gt;)
Only for records. Cross-references another record and makes the other record's data available for the permission rule.
```javascript
{
    "record": {
		//when booking a new car sale, make sure that
		//the car that's sold exists and that its price
		//is the same or lower than what the customer is charged
        "car-sale/$transactionId": {
            "write": "_(data.carId) !== null && _(data.carId).price >= data.price"
        }
    }
}
```

What if multiple rules match my record/event/rpc name?
===================================
Permissioning employs a simple specificity concept based on path lengths.
Rules only overwrite each other for a specific ruletype. Here's an example:
```javascript
{
	// restrictive base rule. Deny everything
	// unless its explicitly allowed by a more specific rule below
	'*': {
		write: false,
		read: false
	},

	// allow users to read their own scores
	// this path is more specific than '*', so it will
	// be applied for read operations. As no write is specified
	// write: false from above will be used
	'score/$userId': {
		read: 'user.id === $userId'
	}
}
```

What about functions?
====================================
Firebase supports a range of functions, e.g. `isString()`, `exists()` etc.. We won't do the same - mainly to make implementation of our permission engine easier and safer and execution faster. Just use `typeof` and any other javascript operaters such as `===`, `>=` etc instead

What happens if my rule errors?
===================================
If your rule is throwing an error, e.g. because you've referenced a non existing field
```javascript
'data.parent.firstname === "Bobo"' //data.parent === undefined
```
it will be treated as false

WolframHempel:
done :-)
FIXES #125


package.json
```diff
@@ -27,7 +32,8 @@
     "istanbul": "^0.4.3",
     "jasmine": "^2.4.1",
     "jasmine-spec-reporter": "^2.4.0",
-    "proxyquire": "1.7.9"
+    "npm-watch": "^0.1.4",
+    "proxyquire": "1.7.4"
```
yasserf: reverted version

src/deepstream.io.js
```diff
@@ -44,7 +44,8 @@ var Deepstream = function() {
 		'messageConnector',
 		'storage',
 		'cache',
-		'logger'
+		'logger',
+		'permissionHandler' //TODO: This now requires the permissionHandler to have a ready flag / emit events
```
yasserf: TODO

yasserf: Lets add a task for milestone 1 to mention this as part of 1.0 release

src/permission/config-permission-handler.js
```diff
@@ -0,0 +1,243 @@
...
+ * Validates and compiles a loaded config. This can be called as the result
+ * of a config being passed to the permissionHandler upon initialisation,
+ * as a result of loadConfig or at runtime
+ *
+ * @todo expose this method via the command line interface
+ *
+ * @param   {Object} config deepstream permissionConfig
+ *
+ * @public
+ * @cli useConfig
```
yasserf: This a tag? Not on usejsdoc

src/permission/config-permission-handler.js
```diff
@@ -0,0 +1,243 @@
...
+ * @param   {Object}   message  a parsed deepstream message
+ * @param   {Function} callback the callback to provide the result
+ * @param   {[Object]}   authData additional optional authData as passed to isValidUser
+ *
+ * @public
+ * @interface
+ * @returns {void}
+ */
+ConfigPermissionHandler.prototype.canPerformAction = function( username, message, callback, authData ) {
+	if( typeof message.data[ 0 ] !== STRING ) {
```
yasserf: If this is so generic, why isn't it in the parser?

src/permission/config-permission-handler.js
```diff
@@ -0,0 +1,243 @@
...
+		callback( 'invalid message', false );
+		return;
+	}
+
+	var ruleSpecification = rulesMap.getRulesForMessage( message );
+	var name = message.data[ 0 ];
+	var ruleData;
+
+	if( ruleSpecification === null ) {
+		callback( null, true );
```
yasserf: Does this not mean it isn't covered by a root permission?

src/permission/config-permission-handler.js
```diff
@@ -0,0 +1,243 @@
...
+	new RuleApplication({
+		username: username,
+		authData: authData,
+		path: ruleData,
+		ruleSpecification:ruleSpecification,
+		message: message,
+		action: ruleSpecification.action,
+		regexp: ruleData.regexp,
+		rule: ruleData.rule,
+		name:name,
```
yasserf: missing space

src/permission/config-permission-handler.js
WolframHempel: We don't check if you have root permissions yet - not sure if we want to enforce them...?

src/permission/config-permission-handler.js
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * @param   {Error} loadError a FileSystem Error that occured during the loading of the file
+ * @param   {String} data     the content of the permission.json file as utf-8 encoded string
+ *
+ * @private
+ * @returns {void}
+ */
+ConfigPermissionHandler.prototype._onConfigLoaded = function( path, loadError, data ) {
+	if( loadError ) {
+		this.emit( 'error', 'error while loading config: ' + loadError.toString() );
```
yasserf: Would this in anyway effect the dependency initializer while hotswapping permissions to something that fails?

src/permission/path-parser.js
```diff
@@ -0,0 +1,61 @@
+var WILDCARD_REGEXP = /\*/g;
+var WILDCARD_STRING = '.*';
+var VARIABLE_REGEXP = /(\$[a-zA-Z0-9]*)/g;
```
yasserf: should this not be a +?

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+
+var C = require( '../constants/constants' );
+var RecordRequest = require( '../record/record-request' );
+var messageParser = require( '../message/message-parser' );
+var JsonPath = require( '../record/json-path' );
+var utils = require( '../utils/utils' );
+
+/**
+ * This class handles the evaluation of a single rule. It creates
+ * the required variables, injects them into the rule function and
```
yasserf: recursively*

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+ * runs the function recoursively until either all cross-references,
```
yasserf: runs the function recursively until either all cross-references and references to old or new data are loaded, when an error occurs or if it exceeds the maximum maxIterationCount

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+ * references to old or new data is loaded, it errors or the maxIterationCount
+ * limit is exceeded
+ *
+ * @constructor
+ *
+ * @param {Object} params requires the following keys
```
yasserf: @param { { username: string, authData: objec  } }

or 
( From usejsdoc )
/**
 * Assign the project to an employee.
 * @param {Object} employee - The employee who is responsible for the project.
 * @param {string} employee.name - The name of the employee.
 * @param {string} employee.department - The employee's department.
 */

etc..

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+	this._user = this._getUser();
+	this._recordData = {};
+	this._id = Math.random().toString();
+	this._iterations = 0;
+	this._run();
+};
+
+/**
+ * Runs the rule function. This method is initially called when this class
+ * is constructed and recoursively from thereon whenever the loading of a record
```
yasserf: recursively

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+		return this._recordData[ this._params.name ];
+	} else {
+		this._loadRecord( this._params.name );
+	}
+};
+
+/**
+ * Compile the list of arguments that will be injected
+ * into the permission function. This method is called
+ * everytime the permission is run. This allows it to merge
```
yasserf: permission <adjective>

permission rule?

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+	}
+
+	var args = this._getArguments();
+	var result;
+
+	if( this._isDestroyed === true ) {
+		return;
+	}
+
+	try{
```
yasserf: TODO ( for yasserf )

src/permission/rule-application.js
```diff
@@ -0,0 +1,420 @@
...
+ *
+ * @param   {String} recordName
+ *
+ * @private
+ * @returns {void}
+ */
+RuleApplication.prototype._loadRecord = function( recordName ) {
+	/* istanbul ignore next */
+	if( this._recordData[ recordName ] === LOADING ) {
+		return;
```
yasserf: worth logging to see if this is ever reached? No way to find out otherwise ( Pretty useless logline though )

src/permission/config-permission-handler.js
yasserf: I think having root permissions is an important part of using the config handler, otherwise you would use the 'open' permission handler instead. Also be weird for like:
```
{
event: {
   '*': {
        ...
   }
}
}
```

Since records would always be true without explicitly mentioning it..

package.json
```diff
@@ -6,7 +6,12 @@
   "directories": {
     "test": "test"
   },
+  "watch": {
```
yasserf: Can we do this the same as the client-js?

"watch": "node node_modules/watch/cli.js \"npm test\" ./src ./test-unit ./test-e2e"

yasserf:
This looks good!

yasserf:
Like real good

yasserf:
Can't run this on my machine properly though =(.

```
	"record": {
		"*": {
			"write": true,
			"read": true
		},
		"someUser" : {
			"read": "data.firstname === \"Yasser\"",
			"write": "data .firstname === \"Yasser\""
		}
	},
```

Gives me:

PERMISSION_ERROR | error when executing function anonymous(_,user,data,oldData,n
ow,action
/**/) {
return data .firstname === "Yasser";
}
for [object Object]: TypeError: Cannot read property 'firstname' of undefined

Also sending the error to the client might be useful for debug purposes, but in production that would most likely be a security concern since clients will be told about the exact permission rule ( minus the error )

Svenskunganka:
You guys are beautiful, this is gonna be awesome! :smile:

WolframHempel:
@yasserf  I've added tests for the case below in [this commit](https://github.com/deepstreamIO/deepstream.io/pull/147/commits/dce8a87517a05f4838205bc50be37c9e6f6dd444) but couldn't reproduce it yet. Would be good to look into this together today 

```javascript
  "record": {
        "*": {
            "write": true,
            "read": true
        },
        "someUser" : {
            "read": "data.firstname === \"Yasser\"",
            "write": "data .firstname === \"Yasser\""
        }
    },
```

src/permission/config-permission-handler.js
WolframHempel: Some messages (e.g. auth `A|A+` / challenge C|CH+ ) don't require data, but they also don't get to canPerformAction. At this point we expect everything to be related to an event/record/rpc/rtc call

src/permission/config-permission-handler.js
WolframHempel: fair enough, I'll just make it a comment to keep track of what we want to expose via the command line interface

package.json
WolframHempel: Happy to change, but doesn't the below feel cleaner?

package.json
WolframHempel: reversion reverted

src/deepstream.io.js
WolframHempel: Raised issue https://github.com/deepstreamIO/deepstream.io-website/issues/69 against website

src/permission/config-permission-handler.js
WolframHempel: Added check for root permissions in [this commit](https://github.com/deepstreamIO/deepstream.io/pull/147/commits/69ca8d75841a101a2d55121bf194a168295b8a58)

src/permission/config-permission-handler.js
WolframHempel: fixed

src/permission/config-permission-handler.js
WolframHempel: If a dependency errors during startup, the dependency initializer will kill the process ([code here](https://github.com/deepstreamIO/deepstream.io/blob/master/src/utils/dependency-initialiser.js#L83))

Once the plugin is ready, its errors are forwarded to the logger ([code here](https://github.com/deepstreamIO/deepstream.io/blob/master/src/deepstream.io.js#L240))

src/permission/path-parser.js
WolframHempel: good spot, updated it

src/permission/rule-application.js
WolframHempel: updated

src/permission/rule-application.js
WolframHempel: british?

src/permission/rule-application.js
WolframHempel: Updated it

src/permission/rule-application.js
WolframHempel: TBH, I don't mind the extra check

package.json
yasserf: could keep it and have the same thing in the client, thought less keys would be cleaner =P

src/permission/rule-application.js
yasserf: Meant due to the comment above 

```
but I'll leave the additional safeguards in until absolutely sure.
```

src/permission/config-permission-handler.js
yasserf: Awesome, thanks for digging that up for me!

src/permission/rule-application.js
yasserf: 4.390 vs 4.800.000

src/permission/config-permission-handler.js
yasserf: Interesting. All A and C messages don't go through to the permission handler anyways since they get directly intercepted within the connection class.

The workflow I see is:
1- [Message recieved from socket after authentication](https://github.com/deepstreamIO/deepstream.io/blob/feature/permission-config/src/message/connection-endpoint.js#L317)
2 - [Which is hooked to the message processor](https://github.com/deepstreamIO/deepstream.io/blob/feature/permission-config/src/deepstream.io.js#L213)
3 - [Which is permissioned](https://github.com/deepstreamIO/deepstream.io/blob/feature/permission-config/src/message/message-processor.js#L56)
4- [And then checked if the type is actually valid](https://github.com/deepstreamIO/deepstream.io/blob/feature/permission-config/src/message/message-distributor.js#L25)

It's important to have that check since it isn't done before, just needed to find out the flow.

src/permission/rule-parser.js
```diff
@@ -0,0 +1,109 @@
+var rulesMap = require( './rules-map' );
+var FUNCTION_REGEXP = /([a-zA-Z0-9_]*)\(/g;
```
yasserf: You could use \w instead:

var FUNCTION_REGEXP = /([\w_]*)\(/g;
var NEW_REGEXP = /^new[^\w^]|[^\w^]new[^\w]/;
var OLD_DATA_REGEXP = /^oldData[^\w^]|[^\w^]oldData[^\w]/;
var DATA_REGEXP = /^data[^\w^]|[^\w^]data[^\w]/;
var CROSS_REFERENCE_REGEXP = /^_[^\w^]|[^\w^]_[^\w]/;

src/permission/rule-parser.js
```diff
@@ -0,0 +1,109 @@
...
+var NEW_REGEXP = /^new[^a-zA-Z0-9^]|[^a-zA-Z0-9^]new[^a-zA-Z0-9]/;
+var OLD_DATA_REGEXP = /^oldData[^a-zA-Z0-9^]|[^a-zA-Z0-9^]oldData[^a-zA-Z0-9]/;
+var DATA_REGEXP = /^data[^a-zA-Z0-9^]|[^a-zA-Z0-9^]data[^a-zA-Z0-9]/;
```
yasserf: Hate to be that person, but '~' technically should also break it

src/permission/rule-parser.js
yasserf: var FUNCTION_REGEXP = /([\w]*)\(/g;
var NEW_REGEXP = /(^|[^\w~])new[^\w~]/;
var OLD_DATA_REGEXP = /(^|[^\w~])oldData[^\w~]/;
var DATA_REGEXP = /(^|[^\w~])data[^\w~]/;
var CROSS_REFERENCE_REGEXP = /(^|[^\w~])_[^\w~]/;

This lets tests pass, reduces complexity and also fixes a bug where '_data and _oldData' was considered legit

yasserf:
This works for everything except:

"( a && b )" would fail since it thinks its a method call

_oldData, _data, ~oldData and ~data slip through the permissions

MessageDenied sometimes returns message raw data sometimes not ( differs from records, events and rpcs )

PR #155, opened 2016-05-19
Made server.set() return itself

prestonkyles:
Allows for a user to do one line configuration:
server.set('host', 'localhost').set('port', '6020');


yasserf:
Makes sense, thanks!

PR #157, opened 2016-05-23
Feature/#145 buildin authentication handler

WolframHempel:


src/authentication/file-based-authentication-handler.js
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * This authentication handler reads a list of users and their associated password (either
+ * hashed or in cleartext ) from a json file. This can be useful to authenticate smaller amounts
+ * of clients with static credentials, e.g. backend provider that write to publicly readable records
+ *
+ * @public
+ * @extends {EventEmitter}
+ */
+module.exports = class FileBasedAuthenticationHandler extends EventEmitter{
```
yasserf: So we moving toward classes in deepstream.io? Not against it, just wasn't aware

src/authentication/file-based-authentication-handler.js
```diff
@@ -0,0 +1,208 @@
...
+
+	/**
+	 * Creates the class, reads and validates the users.json file
+	 *
+	 * @param   {Object} settings
+	 * @param   {String} settings.path path to the user file
+	 * @param   {String} settings.hashAlgo the name of a HMAC digest algorithm, a.g. 'sha512'
+	 * @param   {Int} settings.iterations the amount of times the algorithm should be applied
```
yasserf: Number

src/authentication/file-based-authentication-handler.js
```diff
@@ -0,0 +1,208 @@
...
+
+		var userData = this._data[ authData.username ];
+
+		if( !userData ) {
+			callback( null, false, null );
+			return;
+		}
+
+		if( this._settings.hash ) {
+			this._isValid( authData.password, userData.password, userData.data, callback );
```
yasserf: doesnt this return the password to the user?

src/utils/utils.js
```diff
@@ -9,45 +14,128 @@ exports.getUid = function() {
...
+ *
+ * @param   {Object}  map        the map to validate
+ * @param   {Boolean} throwError if true, errors will be thrown rather than returned
+ * @param   {Object}  schema     json schema in the form { key: type }
+ *
+ * @public
+ * @returns {Boolean|Error}
+ */
+exports.validateMap = function( map, throwError, schema ) {
+	var error, key;
```
yasserf: This is very cool

src/authentication/file-based-authentication-handler.js
yasserf: This is probably the only thing I think may be wierd, also might be good to have a json file with:

{ 
username: string
password: hash || string
data: userdata
}

PR #168, opened 2016-05-26
Feature/recoursive merge

WolframHempel:
Sorry for being annoying :-)

### DIFF
--- a/src/utils/config-loader.js
+++ b/src/utils/config-loader.js
@@ -1,0 +1,82 @@
+var fs = require( 'fs' );
+var path = require( 'path' );
+var yaml = require( 'js-yaml' );
+var merge  = require('lodash.merge');
+var defaultOptions = require( '../default-options' );
+var utils = require( './utils' );
+var  C = require( '../constants/constants' );
+var argv = require( 'minimist' )( process.argv.slice(2) );
+
+module.exports = function(customFilePath) {
+  var order = [
+    'config.json',
+    'config.js',
+    'config.yml',
+  ];
+  if ( customFilePath != null ) {
+    order.unshift(customFilePath)
+  }
+
+  var filePath = order.filter(function( filePath ) {
+    try {
+      fs.lstatSync(filePath)
+      return true
+    } catch ( err ) {}
+  })[0]
+
+  if ( filePath == null ) {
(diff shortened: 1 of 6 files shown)


timaschew:
Nice!

`null` and `undefined` is not working, so the behavior of lodashs merge is that 
`null` _wins_ and `undefined` is ignored.

src/utils/config-loader.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  var filePath = order.filter(function( filePath ) {
+    try {
+      fs.lstatSync(filePath)
+      return true
+    } catch ( err ) {}
+  })[0]
+
+  if ( filePath == null ) {
```
WolframHempel: the value of [ 0 ] for an empty array returned by filter would be undefined

src/utils/config-loader.js
```diff
@@ -0,0 +1,82 @@
...
+    return defaultOptions.get();
```
WolframHempel: I think it might make sense to throw an error if no file could be found. Otherwise you might thing that your confix.json file is used, but due to typo it isn't

src/utils/config-loader.js
```diff
@@ -0,0 +1,82 @@
...
+  var config = null;
+  var extension = path.extname( filePath )
+  try {
+    if ( extension === '.yml' ) {
+      config = yaml.safeLoad( fs.readFileSync( filePath, 'utf8' ));
+    } else {
+      config = require(path.join( process.cwd(), filePath ))
+    }
+  } catch ( err ) {
+    console.error( err )
```
WolframHempel: why not just `throw err` ?

src/utils/config-loader.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  // CLI arguments
+  var cliArgs = {};
+  var keys = Object.keys(defaultOptions.get())
+  for ( key in keys ) {
+    cliArgs[key] = argv[key] || config[key] || defaultOptions.get()[key]
+  }
+
+  var result = merge({}, defaultOptions.get(), handleMagicProperties(config), cliArgs);
```
WolframHempel: not a fan of my merge function, eh?

PR #169, opened 2016-05-27
Feature/#129 config file support

timaschew:


config.yml
```diff
@@ -0,0 +1,67 @@
...
+logLevel: INFO
+
+# Connectivity
+webServerEnabled: true
+tcpServerEnabled: true
+port: 6020
+host: 0.0.0.0
+tcpPort: 6021
+tcpHost: 0.0.0.0
+httpServer: null
```
yasserf: Does this need to be in the yaml file? The argument would have to be a http object which can't be used via a config file.

config.yml
```diff
@@ -0,0 +1,67 @@
...
+urlPath: /engine.io
+
+# SSL Configuration
+sslKey: null
+sslCert: null
+sslCa: null
+
+# Data Manipulation
+dataTransforms: null
```
yasserf: We might also not want to expose this via the yaml file since it expects JS functions and array structure, meaning whoever uses it will probably be doing it via code

timaschew: Yep, fixed in 161e6df1

src/utils/js-yaml-loader.js
```diff
@@ -0,0 +1,150 @@
...
+const path = require( 'path' );
+const yaml = require( 'js-yaml' );
+const defaultOptions = require( '../default-options' );
+const utils = require( './utils' );
+const C = require( '../constants/constants' );
+const argv = require( 'minimist' )( process.argv.slice( 2 ) );
+const LOG_LEVEL_KEYS = Object.keys( C.LOG_LEVEL );
+
+function readAndParseFile( filePath, callback ) {
+	try {
```
WolframHempel: Unnecessary try / catch - readfile will forward error to callback

src/utils/js-yaml-loader.js
```diff
@@ -0,0 +1,150 @@
+'use strict';
```
WolframHempel: Generally, could we please add jsdoc / comments and get test coverage back to 100%

src/utils/js-yaml-loader.js
```diff
@@ -0,0 +1,150 @@
...
+			}
+		} );
+	} catch ( error ) {
+		return callback ( error );
+	}
+}
+
+function parseFile( filePath, fileContent ) {
+	if ( fileContent == null ) {
+		fileContent = fs.readFileSync( filePath, {encoding: 'utf8'} );
```
WolframHempel: Would be better to have a seperate readFileSync function and have parseFile just parse the file

src/utils/js-yaml-loader.js
```diff
@@ -0,0 +1,150 @@
...
+	const order = [
+		'config.json',
+		'config.js',
+		'config.yml'
+	];
+	let filePath = null;
+
+	if ( customFilePath != null ) {
+		try {
+			fs.lstatSync( customFilePath );
```
WolframHempel: Could we make this async as well?

timaschew:
Added some more test case 
and implement plugin lookup by name (alternative for path) with a naming convention.

Test coverage is 100% again.
Feel free to refactor the async calls.

BTW: Travis seems to get stucked with an old commit or something like that

timaschew:
summary of changes after __tests passing__, 6fde5f1 by @WolframHempel 

https://github.com/deepstreamIO/deepstream.io/compare/32dadc87c4e7d038e1c4c02b5493b6be0eb1412f...26912bff32d9e1c5963460296356bbc99083843c

config.yml
```diff
@@ -49,7 +49,22 @@ maxAuthAttempts: 3
...
+# authType: file
+# path: ./users.json #path to the user file
+# hash: 'md5' #the name of a HMAC digest algorithm
+# iterations: 100 #the number of times the algorithm should be applied
+# keyLength: 32 #the length of the resulting key
+
+# authType: http
+# endpointUrl: https://someurl.com/validateLogin
+# permittedStatusCodes: [ 200 ]
+# requestTimeout: 2000
```
timaschew: I would suggest to put the auth stuff into its own object, even as comment.
This makes really easy so switch between the types by just selecting the lines and use the __toggle comments__ feature

```yaml
#Authentication
auth:
  type: open

  # # or file type
  # type: file
  # # path to the user file
  # path: ./users.json 
  # # the name of a HMAC digest algorithm
  # hash: 'md5'
  # # the number of times the algorithm should be applied
  # iterations: 100 
  # # the length of the resulting key
  # keyLength: 32 
  # 
  # # or http type
  # type: http
  # endpointUrl: https://someurl.com/validateLogin
  # permittedStatusCodes: [ 200 ]
  # requestTimeout: 2000
```

config.yml
```diff
@@ -49,7 +49,22 @@ maxAuthAttempts: 3
...
+
 # Permissioning
+permissionType: config
```
timaschew: Same for this.

BTW: permissionType: config and permissionConfigPath sounds contrary when using them both at the same time or is it WIP?

PR #182, opened 2016-06-13
Fixed eventEmitter leak error

WolframHempel:


ronag:
This looks a bit weird? How can a timeout setting fix a leak?

WolframHempel:
Good question: When multiple deepstream clients can provide the same rpc, the remote-rpc-provider-registry chooses the first provider at random to achieve basic load balancing. This behavior is tested [here](https://github.com/deepstreamIO/deepstream.io/blob/master/test/rpc/remote-rpc-provider-registrySpec.js#L136). To determine if a remote provider is available, the provider registry checks how long ago it was registered ([here](https://github.com/deepstreamIO/deepstream.io/blob/master/src/rpc/remote-rpc-provider-registry.js#L156))

The expiration time for remote providers was set to 10 milliseconds. The first provider (privateTopicA) is registered [here](https://github.com/deepstreamIO/deepstream.io/blob/master/test/rpc/remote-rpc-provider-registrySpec.js#L36), the second (privateTopicC)[here](https://github.com/deepstreamIO/deepstream.io/blob/master/test/rpc/remote-rpc-provider-registrySpec.js#L110). Jasmine runs these tests in an asynchronous queue. If more than 10 milliseconds pass since the first provider was registered, `isUpToDate()` returns false, the topic is removed, `hadTopicA` is never true and the while loop keeps being executed. On every execution the server adds the callback as a listener for an incoming rpc provider message until the event emitter limit is breached

ronag:
I see. Then I would say this PR feels more like a workaround than a fix... We should at least check the arguments are valid otherwise the user could provide their own arguments that would strangely cause leaks...

WolframHempel:
Could you describe how the user would  provide their own arguments that would strangely cause leaks?

ronag:
config.yml
```

rpcProviderQueryTimeout: 1000
rpcProviderCacheTime: 1000
```

ronag:
Never mind. I misunderstood.

PR #198, opened 2016-06-17
Adding ability to set important options via cli

yasserf:


yasserf:
@timaschew

bin/deepstream-start
```diff
@@ -70,3 +85,49 @@ if ( program.detach ) {
...
+	}
+	return portNumber;
+}
+
+/**
+* Used by commander to parse the disabling/enabling of servers and fails if invalid
+* value is passed in
+* @private
+*/
+function parseServerEnabled( enabled ) {
```
timaschew: sounds like parseBoolean, I think the name is the second argument, so you could just write

```js
function parseServerEnabled( enabled, name ) {
  ...
  console.error( 'Invalid argument for ' + name + ', please provide true or false' );
}
```

PR #234, opened 2016-06-28
new logo

WolframHempel:
image


src/deepstream.io.js
```diff
@@ -255,18 +257,11 @@ Deepstream.prototype._showStartLogo = function() {
...
-	'                       /                                                      \n'+
-	'                      /                                                       \n'+
-	'=============================== STARTING... ==================================\n';
+	var logo = fs.readFileSync( './ascii-logo.txt', 'utf8' );
+
 	/* istanbul ignore next */
-	process.stdout.write( this._options.colors ? logo.yellow : logo );
+	process.stdout.write( logo + EOL );
+	process.stdout.write( ' =========================   ' + ( this._options.colors ? 'starting'.cyan : 'starting' ) + '   ==========================' + EOL );
 };
```
timaschew: Since the refactoring for the new logger API we dropped the color option at the root level.
Color can be set as a property of the logger options and it got renamed to `colorize`, this is because you can define multiple loggers and we're just passing the whole optoins object to the logger implementation which is winston currently.
The color was inconsistently anyway, because the setColor function was not part of the logger API.

yasserf: Does this mean the colors option here only applies to the logo? Talk about a customizable server! =P

Liking the logo! Missing the 'logo' tho =(

timaschew: Yes, it applies only to the logo.
Definition of the color value is done at 2 placed in the code currently
1: https://github.com/deepstreamIO/deepstream.io/blob/d9e751d26b4e067e57c09ebbfa38cb12b944eeb5/src/default-options.js#L10
2: https://github.com/deepstreamIO/deepstream.io/blob/a734d505a35cc6f9bc45e6ff2990a041b971f3c4/bin/deepstream-start.js#L30

We could introduce a public API to set the color and set it [here like the level](https://github.com/deepstreamIO/deepstream.io/blob/b24a0f6a9b3c142a74ce4c0603eaa5ae27ca68ba/src/config/config-initialiser.js#L93-L97) but this makes only sense for a console type/transport, so not sure if it's really worth.

yasserf: Maybe logo should be once log starts. Bit wierd when using only a file logger if deepstream only shows the logo in console.

Only time that isn't the case is any fails before launch

timaschew: I suggest to omit colors for the logo and keep it at the very first output, like it's already the case.
And color would be dropped at root level

PR #265, opened 2016-07-05
AuthenticationHandler not plugin, dependency plugins don't specifiy t…

yasserf:
@WolframHempel @timaschew 

I was going through the documentation and testing some things, and the valve permissions never actually printed where they were being loaded from.

I know this is way too close to the 1.0 release, but I tested this and it works across all authentication handlers ( minor change ) since all of our connectors isReady events are sync ( file loads so fast it worked regardless ), and also works if they are async.

I think this should go in, but I want to get some thumbs up first incase it:
1) Isn't urgent enough to bypass code freeze
2) Has any implications I can't think of

### DIFF
--- a/src/authentication/open-authentication-handler.js
+++ b/src/authentication/open-authentication-handler.js
@@ -1,4 +1,5 @@
 'use strict';
+const EventEmitter = require( 'events' ).EventEmitter;
 
 /**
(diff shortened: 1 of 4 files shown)


timaschew:
Good catch!
LGTM 👍

Issue #243, opened 2016-06-29
dependency-initialiser calls process.exit() 


mvberg:
If a plugin [fails to initialize](https://github.com/deepstreamIO/deepstream.io/blob/master/src/utils/dependency-initialiser.js#L63-L89) on server.start(), for example the Redis cache connector cannot connect to it's host - the entire process is exited - this seems pretty extreme - thoughts?

WolframHempel:
In my mind that makes sense :-). By configuring a redis connector, you say that you want your data to be persisted. If deepstream can't connect to it, it knows that it wont be able to store your data and everything will be lost if the server is restarted. 
At that point, it doesn't make much sense to continue. Let's rather shut down and fix the problem.

ronag:
Related to #207

timaschew:
@mvberg 
I totally agree with you that to exit the whole process is not a good idea, at least for the usage via an API.

In the new release of deepstream (1.0.0) we provide a CLI. To exit the process in that case makes sense for me. But for the API we will consider to not exit the process, instead to just throw an exception.
So the user can catch it and for instance fix the issue and try to `server.start()` again.

mvberg:
Thank you guys for the feedback.

@timaschew Yep, this is exactly my use case, deepstream isn't the only part of my setup on the server side.

@WolframHempel 

> At that point, it doesn't make much sense to continue. Let's rather shut down and fix the problem.

If you comment out the exit calls, deepstream actually recovers very nicely, firing all ready events (tested with redis-message and redis-cache) and starts itself up when redis comes back online :)

yasserf:
Problem is though that if just commenting out the lines would result in deepstream potentially just hanging, hoping to connect at some point in the future.

Ideally you would be using a cache/db/msg cluster with more redundant entrypoints and not a single point of failure, and hence deepstream would be able to connect to one of them within the time allocated.

You could also disable the functionality but increasing the dependency timeout to something huge, which would avoid exiting. It doesn't solve the error issue, but if an error occurs during a connection is means something is probably wrong enough to deserve a process stop/restart.

The part that I totally agree with is that exiting a process for a node developer must be confusing since you can't actually deal with anything in your own code. It might be worth us bubbling an error up to our cli and exiting there, and allow deepstream users to catch the error and deal accordingly as well

PR #271, opened 2016-07-08
Create CHANGELOG.md

timaschew:


CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
...
+If you used deepstream 0.x before you may need to change some API calls.
+
+###### Permission Handler
+
+In 0.x you can set a `permissionHandler` which needs to implement two functions:
+
+- `isValidUser(connectionData, authData, callback)`
+- `canPerformAction(username, message, callback)`
+
+In deepstream 1.0 the `canPerformAction` is no longer part of the `permissionHandler` instead
```
ronag: `isValidUser` not `canPerformAction`

CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
...
+- otherwise (`isReady` is false) it needs to emit an `ready` event
+- for errors it should emit an `error` event
+
+###### Logger and colors options
+
+The color flag can't be set in the root level of the configuration anymore.
+The default logger will print logs to the StdOut/StdErr in colors.
+You can use the [deepstream.io-logger-winston](https://www.npmjs.com/package/deepstream.io-logger-winston) which can be configured in the config.yml file with several options.
+
+
```
ronag: I think how the configuration options are passed and set is different as well? Before it was `ds.set` now you pass it in the constructor.

timaschew:
NOTE: We should try to stick to a changelog format like http://keepachangelog.com/

yasserf:
I'm more in favor of using the same format as react
https://github.com/facebook/react/blob/master/CHANGELOG.md

yasserf:
## [x.x.x] - yyyy-mm-dd

Release type: [ bug fixes, features, miscellaneous( dependency upgrade, build changes, etc ) ]

### Bug Fixes

- description ( @outsidecontributer #pullrequest )

### Features

- description ( @outsidecontributer #pullrequest )

### Breaking Changes

- description ( @outsidecontributer #pullrequest )

### Removed Deprecations

- description ( @outsidecontributer #pullrequest )

### New Deprecations

- description ( @outsidecontributer #pullrequest )

### Miscellaneous

- description ( @outsidecontributer #pullrequest )

### Contributers

- Name @outsidecontributer

yasserf:
## [1.0.0] - 2016-07-09

### Features

###### CLI
You can start deepstream via a command line interface. You find it in the _bin_ directory. It provides these subcommands:
  - `start`
  - `stop`
  - `status`
  - `install`
  - `info`
  - `hash`
  append a `--help` to see the usage.

###### File based configuration
You can now use a file based configuration instead of setting options via `ds.set(key, value)`.
deepstream is shipped with a _conf_ directory which contains three files:
  - __config.yml__ this is the main config file, you can specify most of the deepstream options in that file
  - __permissions.yml__ this file can be consumed by the PermissionHandler. It's not used by default, but you can enable it in the _config.yml_
  - __users.yml__ this file can be consumed by the AuthenticationHandler. It's not used by default, but you can enable it in the _config.yml_

For all config types support these file types: __.yml__, __.json__ and __.js__

###### Constructor API. 

There are different options what you can pass:
  - not passing any arguments ( consistent with 0.x )
  - passing `null` will result in loading the default configuration file in the directory _conf/config.yml_
  - passing a string which is a path to a configuration file, supported formats: __.yml__, __.json__ and __.js__
  - passing an object which defines several options, all other options will be merged from deepstream's default values

###### Valve permissions rules
You can write your permission into a structured file. This file supports a special syntax, which allows you to do advanced permission checks. This syntax is called __Valve__.

#### Enhancements

###### no process.exit on plugin initialization error or timeout
deepstream will not longer stops your process via `process.exit()`. This happened before when a connector failed to initialise correctly [#243](https://github.com/deepstreamIO/deepstream.io/issues/243) instead it will throw an error now.

Currently the API provides no event or callback to handle this error
other than subscribing to the global `uncaughtException` event.

```javascript
process.once('uncaughtException', err => {
  // err.code will equal to of these constant values:
  // C.EVENT.PLUGIN_INITIALIZATION_TIMEOUT
  // or C.EVENT.PLUGIN_INITIALIZATION_ERROR
})
```
Keep in mind that deepstream will be in an unpredictable state and you should consider to create a new instance.

### Breaking Changes

###### Permission Handler
In 0.x you can set a `permissionHandler` which needs to implement two functions:

- `isValidUser(connectionData, authData, callback)`
- `canPerformAction(username, message, callback)`

In deepstream 1.0 the `isValidUser` and onUserDisconnect`` is no longer part of the `permissionHandler` and is instead within a new handler: `authenticationHandler`.

You can reuse the same 0.x permission handler except you will have to set it on both explicitly.

```javascript
const permissionHandler = new CustomPermissionHandler()
ds.set( 'permissionHandler', permissionHandler )
ds.set( 'authenticationHandler', permissionHandler )
```

###### Plugin API 
All connectors including, the `permissionHandler`, `authenticationHandler` and `logger` all need to implement the plugin interface which means exporting an object that:

- has a constructor
- has an `isReady` property which is true once the connector has been initialized. For example in the case a database connector this would only be `true` once the connection has been established. If the connector is synchronous you can set this to true within the constructor. 
- extends the EventEmitter, and emits a `ready` event once initialized and `error` on error.

###### Logger and colors options
The color flag can't be set in the root level of the configuration anymore.
The default logger will print logs to the StdOut/StdErr in colors.
You can use the [deepstream.io-logger-winston](https://www.npmjs.com/package/deepstream.io-logger-winston) which can be configured in the config.yml file with several options.

###### Connection redirects
deepstream clients now have a handshake protocol which allows them to be redirected to the most efficient node and expect an initial connection ack before logging in. As such In order to connect a client to deepstream server you need also to have a client with version 1.0 or higher.

More details in the [client changelog](https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/CHANGELOG.md).

PR #295, opened 2016-07-19
Fixed remaining subscriber bug

WolframHempel:


yasserf:
What side effect did this have?

WolframHempel:
Side effect: Since we've passed the actual array of subscribers around, it removed subscribers from it in each iteration of unsubscribe. Thus leaving the last one behind

PR #369, opened 2016-10-04
Try to load config file if no params was passed to server constructor

farwayer:
Previous strategy was to load default config. Check config type also.

```javascript
// previous
const server = new Deepstream(); // load defaults
const server = new Deepstream(null); // try to load config file; throw exception if not found
const server = new Deepstream({}); // load defaults
const server = new Deepstream({port: 1234}); // extend defaults

// new
const server = new Deepstream(); // try to load config file; fallback to defaults
const server = new Deepstream(null); // try to load config file; fallback to defaults
const server = new Deepstream({}); // load defaults
const server = new Deepstream({port: 1234}); // extend defaults
```

### DIFF
--- a/src/deepstream.io.js
+++ b/src/deepstream.io.js
@@ -33,7 +33,7 @@ const STATES = C.STATES;
  * @copyright 2016 deepstreamHub GmbH
  * @author deepstreamHub GmbH
  *
- * @param {Object} config Configuration object
+ * @param {Object} [config] Configuration object
  *
  * @constructor
  */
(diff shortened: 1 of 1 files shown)


timaschew:
It was intentionally to throw an error for this case

`const server = new Deepstream(null);`

This helps to avoid a silent handling of a not found config file.
Otherwise you would need to check from where or what the config file was created.

@farwayer Why this is a problem for you?
Can't you just use `const server = new Deepstream(undefined);` ?

farwayer:
Oh, there is no problem. But I expected that constructor with no params load my config file as well as a command line executable file with no arguments. Deepstream loaded default config and it was surprisingly for me.

On the other hand different behavior with `undefined` and `null` is not very good practice.

Anyway you are right and there is better to know if config file is not presented. And it will be nice to have this behavior documented. I spent some time trying to figure out why my configuration could not be loaded.

Issue #326, opened 2016-08-17
Feature Request - GET /healthcheck


ronag:
It would be nice if deepstream could provide a GET /healthcheck HTTP route for health monitoring third-party services.

Any recommendations on how to implement this in user space?

WolframHempel:
Yup, I'm in favor of having that as well. You can set up most  loadbalancers to run health checks against the TCP port, but if you want to use it with e.g. AWS application ELB, you need a HTTP endpoint that returns 200

PR #433, opened 2016-10-31
Enhancement/health check

jdmnd:
Added health check support
https://github.com/deepstreamIO/deepstream.io/issues/326

### DIFF
--- a/conf/config.yml
+++ b/conf/config.yml
@@ -15,6 +15,8 @@ port: 6020
 host: 0.0.0.0
 # url path websocket connections connect to
 urlPath: /deepstream
+# url path for http health-checks, GET requests to this path will return 200 if deepstream is alive
+healthCheckPath: /health-check
 # webfacing URL under which this client is reachable. Used for loadbalancing / failover
 externalUrl: null
 # the time to wait for a buffer to fill before sending it out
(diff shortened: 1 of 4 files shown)


conf/config.yml
```diff
@@ -15,6 +15,8 @@ port: 6020
 host: 0.0.0.0
 # url path websocket connections connect to
 urlPath: /deepstream
+# url path for http health-checks, GET requests to this path will return 200 if deepstream is alive
+healthCheckPath: /health-check
```
ronag: I think /healthcheck is the most common default?

ricard0javier:
Hello guys, I'm trying to use this new feature but when I hit  `http://localhost:6020/health-check` I don't receive the 200 status, instead the connection is kept. I'm using the example configuration file in https://github.com/deepstreamIO/deepstream.io/blob/b0002c5bff3210866a3e6356a935e0f33832d8aa/conf/config.yml and a docker image to run deepstream. What an I missing here?

ricard0javier:
Ok, I just realised that the changelog in https://github.com/deepstreamIO/deepstream.io/blob/master/CHANGELOG.md does mention that the right endpoint is http://localhost/health-check and it is available in the 2.0.0 version, but the latest version available for docker is 1.1

PR #436, opened 2016-11-03
Improve sendToSubscribers performance

alexhultman:
New try...

### DIFF
--- a/conf/config.yml
+++ b/conf/config.yml
@@ -149,3 +149,5 @@ listenResponseTimeout: 500
 lockTimeout: 1000
 # The amount of time a lock request waits for before it defaults to false
 lockRequestTimeout: 1000
+# The amount of time a broadcast will wait (to allow broadcast coalescing). -1 means disabled.
+broadcastTimeout: 0
--- a/src/default-options.js
+++ b/src/default-options.js
@@ -78,7 +78,8 @@ exports.get = function() {
 		clusterNodeInactiveTimeout: 6000,
 		listenResponseTimeout: 500,
 		lockTimeout: 1000,
-		lockRequestTimeout: 1000
+		lockRequestTimeout: 1000,
+		broadcastTimeout: 0
 	};
 
 	return options;
--- a/src/message/socket-wrapper.js
+++ b/src/message/socket-wrapper.js
@@ -1,7 +1,8 @@
 var C = require( '../constants/constants' ),
 	messageBuilder = require( './message-builder' ),
 	EventEmitter = require( 'events' ).EventEmitter,
-	utils = require( 'util' );
+	utils = require( 'util' ),
+	uws = require( 'uws' );
 
 /**
  * This class wraps around a websocket
@@ -24,6 +25,7 @@ var SocketWrapper = function( socket, options ) {
 	this.authCallBack = null;
 	this.authAttempts = 0;
 	this.setMaxListeners( 0 );
+	this.uuid = Math.random();
 
 	this._queuedMessages = [];
 	this._currentPacketMessageCount = 0;
@@ -40,6 +42,61 @@ var SocketWrapper = function( socket, options ) {
 };
 
 utils.inherits( SocketWrapper, EventEmitter );
+SocketWrapper.lastPreparedMessage = null;
+
+/**
+ * Updates lastPreparedMessage and returns the [uws] prepared message.
+ *
+ * @param {String} message the message to be prepared
+ *
+ * @public
+ * @returns {External} prepared message
+ */
+SocketWrapper.prepareMessage = function( message ) {
+	SocketWrapper.lastPreparedMessage = message;
+	return uws.native.server.prepareMessage( message, uws.OPCODE_TEXT );
+}
+
+/**
+ * Sends the [uws] prepared message, or in case of testing sends the
+ * last prepared message.
+ *
+ * @param {External} preparedMessage the prepared message
+ *
+ * @public
+ * @returns {void}
+ */
+SocketWrapper.prototype.sendPrepared = function(preparedMessage) {
+	if ( this.socket.external ) {
+		uws.native.server.sendPrepared( this.socket.external, preparedMessage );
+	} else if ( this.socket.external !== null ) {
+		this.socket.send( SocketWrapper.lastPreparedMessage );
+	}
+}
+
+/**
+ * Variant of send with no particular checks or appends of message.
+ *
+ * @param {String} message the message to send
+ *
+ * @public
+ * @returns {void}
+ */
+SocketWrapper.prototype.sendNative = function( message ) {
+	this.socket.send( message );
+}
+
+/**
+ * Finalizes the [uws] perpared message.
+ *
+ * @param {External} preparedMessage the prepared message to finalize
+ *
+ * @public
+ * @returns {void}
+ */
+SocketWrapper.finalizeMessage = function( preparedMessage ) {
+	uws.native.server.finalizeMessage( preparedMessage );
+}
 
 /**
  * Returns a map of parameters that were collected
@@ -96,9 +153,8 @@ SocketWrapper.prototype.sendMessage = function( topic, action, data ) {
 };
 
 /**
- * Main method for sending messages. Doesn't send messages instantly,
- * but instead achieves conflation by adding them to the message
- * buffer that will be drained on the next tick
+ * Checks the passed message and appends missing end separator if
+ * needed, and then sends this message immediately.
  *
  * @param   {String} message deepstream message
  *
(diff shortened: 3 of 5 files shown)


ronag:
@alexhultman: Would you mind commenting the code or maybe use more meaningful variable names? Maybe it's just me but I'm having a bit of a hard time following this...

alexhultman:
Yep I just wanted to test it against Travis. Need to get the tests running locally tomorrow. This patch doesn't follow the style of the project and the names will be replaced.

alexhultman:
This will need some more fixing and more work in SocketWrapper but at least tests now pass.

src/message/socket-wrapper.js
```diff
@@ -3,6 +3,14 @@ var C = require( '../constants/constants' ),
 	EventEmitter = require( 'events' ).EventEmitter,
 	utils = require( 'util' );
 
+const uws = ( () => {
```
yasserf: This `if` shouldn't be needed as uws is now baked in

alexhultman:
@WolframHempel @yasserf This commit will (optionally) delay any broadcast using a timer *per subscription*. What this means is that if you emit events of a room (subscription name), these emits will be in order in relation to each other, but not in relation to other emits of different rooms or other sends.

If the server receives an emit of "event1" followed by "event2" and then "event1" again, this will be broadcasted to the listeners over their TCP streams as two DS messages: (event1+event1) and (event2). This reorder is important to allow efficient merges of the broadcast and from what I can see this behavior should not be of any real problem as long as clients think of emits and record updates as totally "async". Aka, they shouldn't rely on some kind of specific order between different rooms.

It is possible to enforce a more strict order but that would require a lot of work in order to not ruin the performance gain itself (and I don't really think strict order should be a priority). Given that this PR improves performance of any broadcast by up to 273x I think it is worth the sacrifice of strict ordering.

src/utils/subscription-registry.js
```diff
@@ -111,18 +117,89 @@ class SubscriptionRegistry {
 	 * @returns {void}
 	 */
 	sendToSubscribers( name, msgString, sender ) {
-		if( this._subscriptions[ name ] === undefined ) {
+		function onTimeout( delayedBroadcasts, that ) {
+			var sockets = that._subscriptions[ delayedBroadcasts.name ];
+			if ( sockets ) {
+				delayedBroadcasts.uniqueSendersVector.sort( (a, b) => {
```
ronag: `(a, b) => a.sender - b.sender`?

src/message/socket-wrapper.js
alexhultman: You don't fall back to ws when used as a Node.js module?

alexhultman:
Okay I think this is done enough now. It skips the former SocketWrapper.send queue and instead relies on the sendToSubscribers improvement. This means you will get the old strictly ordered behavior with broadcastTimeout: -1 set. By default the timeout is 0.

src/default-options.js
```diff
@@ -78,7 +78,8 @@ exports.get = function() {
 		clusterNodeInactiveTimeout: 6000,
 		listenResponseTimeout: 500,
 		lockTimeout: 1000,
-		lockRequestTimeout: 1000
+		lockRequestTimeout: 1000,
+		broadcastTimeout: 0
```
yasserf: this needs to be put in the yaml conf/config file and documentated

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,28 @@ var SocketWrapper = function( socket, options ) {
 };
 
 utils.inherits( SocketWrapper, EventEmitter );
+var lastPreparedMessage;
+
+SocketWrapper.prepareMessage = function( message ) {
```
yasserf: missing documentation

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,28 @@ var SocketWrapper = function( socket, options ) {
 };
 
 utils.inherits( SocketWrapper, EventEmitter );
+var lastPreparedMessage;
```
yasserf: we don't do things like this ( in global scope )

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,28 @@ var SocketWrapper = function( socket, options ) {
...
+
+SocketWrapper.prepareMessage = function( message ) {
+	lastPreparedMessage = message;
+	return uws.native.server.prepareMessage( message, uws.OPCODE_TEXT );
+}
+
+SocketWrapper.prototype.sendPrepared = function(preparedMessage) {
```
yasserf: doc

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,28 @@ var SocketWrapper = function( socket, options ) {
...
+	if ( this.socket.external ) {
+		uws.native.server.sendPrepared( this.socket.external, preparedMessage );
+	} else if ( this.socket.external !== null ) {
+		this.socket.send( lastPreparedMessage );
+	}
+}
+
+SocketWrapper.prototype.sendNative = function( message ) {
```
yasserf: doc

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,28 @@ var SocketWrapper = function( socket, options ) {
...
+	this.socket.send( message );
+}
+
+SocketWrapper.finalizeMessage = function( preparedMessage ) {
```
yasserf: doc

src/message/socket-wrapper.js
```diff
@@ -114,6 +138,11 @@ SocketWrapper.prototype.send = function( message ) {
 		return;
 	}
 
+	// as part of the optimization of sendToSubscribes, we no longer rely
+	// on the former per-socket queue. We now send immediately from here.
+	this.socket.send( message );
+
+	/*
```
yasserf: we don't leave commented out code, thats what git is for

src/utils/subscription-registry.js
```diff
@@ -111,18 +117,94 @@ class SubscriptionRegistry {
 	 * @returns {void}
 	 */
 	sendToSubscribers( name, msgString, sender ) {
-		if( this._subscriptions[ name ] === undefined ) {
+		function onTimeout( delayedBroadcasts, that ) {
```
yasserf: this function needs to be cleaned up, preferebly split into separate functions

src/utils/subscription-registry.js
```diff
@@ -111,18 +117,94 @@ class SubscriptionRegistry {
...
+			delayedBroadcasts.uniqueSendersMap = {};
+			delayedBroadcasts.timer = null;
+		}
+
+		if (this._subscriptions[ name ] == undefined) {
 			return;
 		}
 
-		var i, l = this._subscriptions[ name ].length;
+		// todo: this should not have to be corrected here! fix where the real issue is!
```
yasserf: ?

alexhultman: Some messages are invalid when they are passed to be sent, so some of them get the end char appended here.

yasserf: when would a message passed be invalid? That sounds like a higher level issue

conf/config.yml
```diff
@@ -149,3 +149,5 @@ listenResponseTimeout: 500
 lockTimeout: 1000
 # The amount of time a lock request waits for before it defaults to false
 lockRequestTimeout: 1000
+# The amount of time a broadcast will wait (to allow broadcast coalescing). -1 means disabled.
```
yasserf: 👍

src/utils/subscription-registry.js
```diff
@@ -101,7 +107,76 @@ class SubscriptionRegistry {
...
+				if ( a.sender < b.sender ) {
+					return -1;
+				} else if ( a.sender > b.sender ) {
+					return 1;
+				} else {
+					return 0;
+				}
+			} );
+
+			// for all unique senders and their gaps, build their special messages
```
yasserf: what is a gap?

src/utils/subscription-registry.js
```diff
@@ -101,7 +107,76 @@ class SubscriptionRegistry {
...
+	 * @public
+	 * @returns {void}
+	 */
+	onBroadcastTimeout( delayedBroadcasts ) {
+		let sockets = this._subscriptions[ delayedBroadcasts.name ];
+		if ( sockets ) {
+			// sort vector of unique senders by uuid. doing so in combination with
+			// the sorting of the sockets in this subscription name means we can
+			// simplify comparisons
+			delayedBroadcasts.uniqueSendersVector.sort( (a, b) => {
```
yasserf: comparisons of?

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,53 @@ var SocketWrapper = function( socket, options ) {
 };
 
 utils.inherits( SocketWrapper, EventEmitter );
+SocketWrapper.lastPreparedMessage = null;
```
alexhultman: I mean, it's kind of hard to describe it anymore than what the code and the name does. It's kind of obvious.

src/utils/subscription-registry.js
```diff
@@ -101,7 +107,76 @@ class SubscriptionRegistry {
...
+				for ( let j = 1; j < uniqueSender.gaps.length; j++ ) {
+					uniqueSender.message += delayedBroadcasts.sharedMessages.
+						substring( lastStop, uniqueSender.gaps[ j ].start );
+					lastStop = uniqueSender.gaps[ j ].stop;
+				}
+				uniqueSender.message += delayedBroadcasts.sharedMessages.
+					substring( lastStop, delayedBroadcasts.sharedMessages.length );
+			}
+
+			// for all sockets in this subscription name, send either sharedMessage or this socket's
```
yasserf: 👍

src/message/socket-wrapper.js
alexhultman: The comment of prepareMessage even states it.

src/message/socket-wrapper.js
```diff
@@ -40,6 +42,53 @@ var SocketWrapper = function( socket, options ) {
...
+
+/**
+ * Sends the [uws] prepared message, or in case of testing sends the
+ * last prepared message.
+ *
+ * @public
+ * @returns {void}
+ */
+SocketWrapper.prototype.sendPrepared = function(preparedMessage) {
+	if ( this.socket.external ) {
```
alexhultman: socket is an uws socket so external is internal to uws. This is how you check if the socket's external can be used (external is v8::External).

yasserf:
@alexhultman  @WolframHempel I'm happy with this, tests are lacking for new functionality but I'm not really sure if we want it on a UT layer as much as ensuring the e2e tests pass on the client. Benchmarks also test these cases in depth and guarantee tests pass with 1->10k users and 1-10k messages

src/message/socket-wrapper.js
alexhultman: First check is if this socket is valid aka not disconnected. Then the next is a check if external is not null (which means it is undefined instead of null which means it is not an uws socket)

src/message/socket-wrapper.js
yasserf: > Updates lastPreparedMessage and returns the [uws] prepared message.

And a uws prepared message is? Why is it static? What does it do? In nodeJS this concept is not as well known and since it calls into uws i would argue that it isn't obvious at all why its being used.

Same thing for finalizeMessage. Why is it done after sending the message? It is a flush? Is it a delete?

src/message/socket-wrapper.js
yasserf: So its a code branch specifically written for test scenarios. Does this only matter in the UTs for socket wrapper? everything else should use a mock socket wrapper anyways. Using test codepaths in code is generally confusing and not something we have anywhere in any of our repos.

PR #444, opened 2016-11-15
[WIP] Presence fix

AlexBHarley:
- presence permissions

- handling unsubscribes

### DIFF
--- a/src/permission/config-schema.js
+++ b/src/permission/config-schema.js
@@ -20,5 +20,10 @@ module.exports = {
 	rpc: {
 		provide: true,
 		request: true
+	},
+	presence: {
+		query: true,
+		logins: true,
+		logouts: true
 	}
 };
--- a/src/permission/rules-map.js
+++ b/src/permission/rules-map.js
@@ -20,6 +20,7 @@ RULE_TYPES.PUBLISH = 	{ name: 'publish', 		data: true, 	oldData: false };
 RULE_TYPES.SUBSCRIBE = 	{ name: 'subscribe', 	data: true, 	oldData: false };
 RULE_TYPES.PROVIDE = 	{ name: 'provide', 		data: false, 	oldData: false };
 RULE_TYPES.REQUEST = 	{ name: 'request', 		data: true, 	oldData: false };
+RULE_TYPES.QUERY = 		{ name: 'query', 		data: false, 	oldData: false };
 
 /**
  * This class maps topic / action combinations to applicable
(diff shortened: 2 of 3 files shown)


src/permission/rules-map.js
```diff
@@ -60,6 +61,14 @@ RULES_MAP[ C.TOPIC.RPC ].actions = {};
 RULES_MAP[ C.TOPIC.RPC ].actions[ C.ACTIONS.SUBSCRIBE ] = RULE_TYPES.PROVIDE;
 RULES_MAP[ C.TOPIC.RPC ].actions[ C.ACTIONS.REQUEST ] = RULE_TYPES.REQUEST;
 
+RULES_MAP[ C.TOPIC.PRESENCE ] = {};
+RULES_MAP[ C.TOPIC.PRESENCE ].section = 'presence';
+RULES_MAP[ C.TOPIC.PRESENCE ].actions = {};
+RULES_MAP[ C.TOPIC.PRESENCE ].actions[ C.ACTIONS.SUBSCRIBE ] = RULE_TYPES.SUBSCRIBE;
+RULES_MAP[ C.TOPIC.PRESENCE ].actions[ C.ACTIONS.SUBSCRIBE ] = RULE_TYPES.SUBSCRIBE;
```
AlexBHarley: At the moment, we subscribe to both login and logout events. How to fine tune permissions to allow subscriptions to just login or just logout events - or do we even need to worry about that?

src/permission/config-schema.js
```diff
@@ -20,5 +20,10 @@ module.exports = {
 	rpc: {
 		provide: true,
 		request: true
+	},
+	presence: {
+		query: true,
+		logins: true,
+		logouts: true
```
AlexBHarley: Or should we just have

presence: {
` ` query: true,
` ` sessions: true, //where sessions is both logins and logouts
}

yasserf: QUERY
SUBSCRIBE

be good to keep with terminology to make it easier

src/permission/rules-map.js
yasserf: no, if your permission for just subscribe that should be enough.

src/presence/presence-handler.js
```diff
@@ -120,6 +120,31 @@ module.exports = class PresenceHandler {
...
+	 * Handles unsubscriptions to client login and logout events
+	 *
+	 * @param   {Object} socketWrapper the socketWrapper of the client that left
+	 * @param   {String} actions either a presence leave or join message
+	 *
+	 * @private
+	 * @returns {void}
+	 */
+	_handleUnsubscribe( socketWrapper, action ) {
+		if( action === C.ACTIONS.PRESENCE_JOIN ) {
```
yasserf: i would argue that the front end would only do a subscribe, which means it would receive join and leave events, but it doesn't subscribe to each individually. Seems like an odd case to only be permission to be told when people leave.

AlexBHarley:
Closing in favour of #446

PR #451, opened 2016-11-21
Use uWS api to retreive handshakeData

### LINKED ISSUE #450: Missing connectionData with 2.0.0
Just upgraded to deepstream 2.0.0 and noticed that connectionData sent to http auth via POST is now blank (connectionData: {}). Previously I could retrieve connectionData.remoteAddress (IP address of the client logging in). Is this a bug?

jdmnd:
Fixes #450


jdmnd:
Pushed a broken commit, will add fix shortly!

PR #462, opened 2016-11-30
Adding additional tests for Valve parsing and fixing most of them

Iiridayn:
Not sure whether fixing `".*new.*"` matters or not, can be commented out if we don't care; making a case I couldn't break would probably require a JS parser. I _very probably_ haven't caught all of the ways a function could be called, just the ones that were obvious to me.

### DIFF
--- a/src/permission/rule-parser.js
+++ b/src/permission/rule-parser.js
@@ -1,7 +1,7 @@
 var rulesMap = require( './rules-map' );
 
 // TODO: any of these are fine inside a string or comment context...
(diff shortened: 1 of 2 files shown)


Iiridayn:
Yep, missed at least one more obvious way to call a function added in ES6:
```
console.log`foo`
```

yasserf:
Closing since I branched of your branch to match eslint. There was a single failing test I had to comment out where it returns a false negative with 'new' string value

PR #464, opened 2016-12-01
Auth data fixes

### LINKED ISSUE #463: Server error for invalid messages during challenge / auth sequence
Received :  C|CH+
Sending :  ||+

[1:22]  
TypeError: Cannot read property 'topic' of undefined
TypeError: Cannot read property 'topic' of undefined

[1:22]  
at ConnectionEndpoint._processConnectionMessage (nexe.js:24329:14)
   at Server.native.server.group.onMessage (nexe.js:1225:23)

AlexBHarley:
Fixes #463

### DIFF
--- a/src/message/connection-endpoint.js
+++ b/src/message/connection-endpoint.js
@@ -191,7 +191,7 @@ ConnectionEndpoint.prototype._onConnection = function( socket ) {
 ConnectionEndpoint.prototype._processConnectionMessage = function( socketWrapper, connectionMessage ) {
 	var msg = messageParser.parse( connectionMessage )[ 0 ];
 
-	if( msg === null ) {
+	if( msg === null || msg === undefined ) {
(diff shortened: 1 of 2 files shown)


src/message/connection-endpoint.js
```diff
@@ -488,4 +488,20 @@ ConnectionEndpoint.prototype._isHttpsServer = function( ) {
...
+* Checks for authentication data and throws if null or not well formed
+*
+* @throws Will throw an error on invalid auth data
+*
+* @private
+* @returns {void}
+*/
+ConnectionEndpoint.prototype._getValidAuthData = function( authData ) {
+	var parsedData = JSON.parse( authData );
+	if( parsedData === null || parsedData === undefined ) {
```
yasserf: Should we add a type object check as  well? Otherwise you can login with a number or string which would be a bit wierd

PR #467, opened 2016-12-06
adding setup handshake data and tests

### LINKED ISSUE #450: Missing connectionData with 2.0.0

AlexBHarley:
Fixes #450.

Previously handshake data was overwritten with every call to getHandshakeData().


yasserf:
Very nice!

PR #486, opened 2016-12-11
Enhancement/stricter valve checks

yasserf:


Iiridayn:
Continuing on from #462, I missed another fun way to call a function:

`var foo['/foo/'] = console.log; foo[/foo/]('test');`

I can't remember offhand why I didn't make `FUNCTION_REGEXP = /([\w]+\]?)\s*\(/g` - probably an oversight, since `var log = 'log'; console[log]('test');` would also execute.

PR #490, opened 2016-12-13
Relative file path for config

AlexBHarley:


src/config/js-yaml-loader.js
```diff
@@ -189,6 +189,7 @@ function getDefaultConfigPath() {
   let filePath
   let i
   let k
+  DEFAULT_CONFIG_DIRS = process.env.configDirs || DEFAULT_CONFIG_DIRS
```
yasserf: i would keep it as a const and use a different variable, things in the 'global' scope shouldn't be let if possible

PR #500, opened 2016-12-15
Replace os.tmpDir() with os.tmpdir()

ChristophConrads-deepstreamHub:
These functions are aliases:
https://github.com/nodejs/node/blob/4a5c719be50c7998684768b54d06860cb2c1d6fb/lib/os.js#L54-L55

This silences a warning when starting `bin/deepstream`:
`(node:1224) DeprecationWarning: os.tmpDir() is deprecated. Use os.tmpdir() instead.`

### DIFF
--- a/bin/pid-helper.js
+++ b/bin/pid-helper.js
@@ -5,7 +5,7 @@ const fs = require('fs')
 const path = require('path')
 const os = require('os')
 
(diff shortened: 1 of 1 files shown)


yasserf:
Interesting, what version of nice after you using?

PR #537, opened 2017-01-19
Fix #533 using a callback

### LINKED ISSUE #533: `delete` events don't flow to clients connected before mine
Common setup:
1. load browser window 1
2. load browser window 2
3. In both (order doesn't matter): `var r = client.record.getRecord('test'); r.on('delete', function() { console.log('deleted'); });`

This is incorrect:
1. Run common setup
2. In 2: `r.delete()`
3. Notice how 1 doesn't show the deleted message
4. In 1: `r.delete()`
5. Now it does, though the record wasn't in the database anyway

It should work the same as:
1. Run common setup
2. In 1: `r.delete()`
3. Notice how 2 also shows the deleted message
4. In 2: `r.delete()`
5. Notice this time deepstream complains that the record is already destroyed

mclark-newvistas:


yasserf:
Hey, closing this since I'm not sure how the callback is being passed through. Will certainly use it as a baseline when we actually look at #533. Thanks for raising it!

PR #554, opened 2017-02-06
message-processor: Optimized + Ensure message ordering

ronag:
This fixes several edge cases where messages going through the permission handler can be re-ordered... e.g. a `UNSUBSCRIBE` could be handled before `READ` even though the client sent `READ` first. Same problem in other cases such as `UPDATE` and  `PATCH`.

Message ordering for a specific client should be maintained.

### DIFF
--- a/src/message/connection-endpoint.js
+++ b/src/message/connection-endpoint.js
@@ -203,25 +203,25 @@ module.exports = class ConnectionEndpoint extends events.EventEmitter {
    * @returns {void}
    */
   _processConnectionMessage(socketWrapper, connectionMessage) {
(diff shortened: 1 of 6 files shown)


ronag:
This should have minimal performance impact with some extra memory overhead.

ronag:
Build fails on node 6.9.1? That's a bit strange.

ronag:
@yasserf: We've merged this into our fork without problems so far.

yasserf:
This makes alot of sense, raised https://github.com/deepstreamIO/deepstream.io/pull/610

Looks like a few new things need to be ported (mainly ignoring non text messages).

PR #558, opened 2017-02-07
message-parser: Optimized

ronag:
Optimizes message parsing by avoiding extra array allocation and iteration, taking advantage of the min 3 char length of messages and making the parsing a bit more cpu cache friendly.


yasserf:
Do you have any micro benchmarks of the difference this makes?

ronag:
@yasserf: nope, waiting for you benchmark framework.

ronag:
@yasserf: It's actually significantly slower. Will close this as nothing but a fun experiment. I need to stop thinking javascript is C/C++.

PR #569, opened 2017-02-11
rpc-handler: Fixed getAlternativeProvider. Fixes #566

### LINKED ISSUE #566: RpcHandler.getAlternativeProvider can return invalid provider?
Since `getAlternativeProvider` uses a copy of subscribers (`local`) https://github.com/deepstreamIO/deepstream.io/blob/master/src/rpc/rpc-handler.js#L216, which is not updated when a provider is unsubscribed or closes, it can return an invalid/closed/unsubscribed provider?

@yasserf: bug?

ronag:


yasserf:
Looks good and tests are happy, thanks!

PR #575, opened 2017-02-12
Listening bug

yasserf:
@ronag @jdmnd

### DIFF
--- a/src/listen/listener-registry.js
+++ b/src/listen/listener-registry.js
@@ -205,11 +205,9 @@ module.exports = class ListenerRegistry {
       this._sendHasProviderUpdateToSingleSubscriber(true, socketWrapper, subscriptionName)
       return
     }
-
     if (localCount > 1) {
       return
     }
-
     this._startDiscoveryStage(subscriptionName)
   }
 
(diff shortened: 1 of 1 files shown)


ronag:
I'm not much help here. The listening part is unfortunately the only code that I don't understand in deepstream...

yasserf:
Are you okay with the decision I made and stated in the issue regarding unsubscribing the client from the server?

yasserf:
Should also make a note to draw a nice big complicate state diagram for listening lol

ronag:
> Are you okay with the decision I made and stated in the issue regarding unsubscribing the client from the server?

I don't know. Maybe I'm to tired right now. I'll try to think about it again next week.

ronag:
This PR seems to lead to lots of weird `NOT_SUBSCRIBED` errors... debugging...

ronag:
Yep, doesn't happen after reverting...

ronag:
@yasserf: What can I do to help with this?

yasserf:
Will need to investigate further, are you discarding again on your app?

yasserf:
A solution is to discard records your providing when subscription of lost, just gets duplicated easy in application logic

ronag:
@yasserf: Sorry... I didn't understand any of that? No my app only discards once, I've got invariants making sure of that. If it did it twice I would have had the error even without this PR, no?

yasserf:
Closing due to issues.

> @yasserf: Sorry... I didn't understand any of that? No my app only discards once, I've got invariants making sure of that. If it did it twice I would have had the error even without this PR, no?

What I meant is that since the server discards it in this PR your discard would be considered unsolicited.

ronag:
> What I meant is that since the server discards it in this PR your discard would be considered unsolicited.

Ah, now I understand... maybe that's fine... but it's a bit weird and does feel a bit like a workaround/hack...

PR #588, opened 2017-02-16
Listener registry re-write.

ronag:
This is a work in progress. I've opened this PR mostly for discussions and comments.

The basic idea is to simply use an atomic shared state to handle listeners, i.e. no message connector required and implicit shared state. This also allows the implementation to be "self-healing".

Also, I have removed the handling of the case where the provider also is subscribing. Instead I suggest that users should be directed to use a new helper method inside the provider when updating the matched record, e.g:

```js
ds.record.set = function (name, ...args) {
  const record = this.getRecord(name, { weak: true })
  record.set(name, ...args)
  record.discard()
}
```

For the case where the user incorrect subscribes we could have the server reset the provider state for a specific record name every minute or so, recover from the problem and print a warning.

There is a lot that could be optimized here but I'm striving to keep it as simple and correct as possible. So that there is a fully stable baseline.


ronag:
For the case where the user incorrect subscribes we could have the server reset the provider state for a specific record name every minute or so, recover from the problem and print a warning.

ronag:
@yasserf: This is starting to be pretty solid in theory. Haven't run any test against it yet. If you have any comments or suggestion those would be appreciated before I take this further.

ronag:
I'm going to need some updates to the distributed state registry in order to get this working with the current cluster implementation... in particular I need:

```
state.watch(table, key, (err, name, value) => { ... })
state.keys(table)
state.upsert(table, key, value => { ... }).then(([ prev, next ]) => { ... }).catch(err => { ... })
```

yasserf:
I have the distributed state registry connector up next on my list, hoping to be able to do the rewrite tomorrow!

ronag:
In order to get this to work we will need weak records. I imagine that "weak" records are basically records that send a `READ` that does NOT subscribe to the record.

```js
const record = ds.record.getRecord('foo')
const weakRecord = ds.record.getRecord('foo', { weak: true })
```

ronag:
Another limitation (which I believe is already the case?) is that the same client CANNOT listen to the same pattern twice.

ronag:
@yasserf: I'm putting further work on this on hold until your team gets involved or we start to get near to deploy in production. It's not urgent for us but we need listeners resolved before we go into production ourselves. Ping me.

yasserf:
Hey Ronag, given the amount of work required I'll be looking at this post 6th march. I have however made progress with the state registry, just need to figure how to reconcile on server crashes

PR #599, opened 2017-03-18
Optimize Subscription Registry

ronag:
Third time is a charm. Passes all unit & e2e tests.

Several misc optimization but basically goes from O(n^2) to O(n).


yasserf:
I think you means charm right? =P

yasserf:
This generally looks great. I will be merging it in but will do an RC release to ensure it works as expected.

Thank you for doing this btw, its a huge performance win! Looking forward to seeing how it will effect performance results

PR #604, opened 2017-03-24
Move record logic around getting record config into utils

AlexBHarley:


src/record/record-transition.js
```diff
@@ -131,7 +132,8 @@ RecordTransition.prototype.add = function (socketWrapper, version, message) {
     }
 
     try {
-      this._applyConfig(update, message)
+      const config = utils.getRecordConfig(message)
```
yasserf: utils is for generic stuff, this would make more sense to just be a static method on this class

PR #605, opened 2017-03-24
RPC opt + cleanup

ronag:


src/listen/listener-registry.js
```diff
@@ -858,8 +858,7 @@ module.exports = class ListenerRegistry {
   * @param   {String} errorMsg
   * @param   {Event} [errorEvent] Default to C.EVENT.INVALID_MESSAGE_DATA
   */
-  _onMsgDataError (socketWrapper, errorMsg, errorEvent) {
-    errorEvent = errorEvent || C.EVENT.INVALID_MESSAGE_DATA // eslint-disable-line
+  _onMsgDataError (socketWrapper, errorMsg, errorEvent = C.EVENT.INVALID_MESSAGE_DATA) {
```
yasserf: this isn't Node 4 compatible

src/rpc/rpc-handler.js
```diff
@@ -320,13 +294,13 @@ module.exports = class RpcHandler {
   * @private
   * @returns {Boolean} isValid
   */
- // eslint-disable-next-line
   _isValidMessage (dataLength, socketWrapper, message) {
```
yasserf: this is needed because `this` is not used in the method

src/rpc/rpc-handler.js
```diff
@@ -335,7 +309,6 @@ module.exports = class RpcHandler {
   *
   * @return {String}
   */
- // eslint-disable-next-line
```
yasserf: this is needed because this is not used in the method

we can make it static if needed instead

yasserf:
Made changes in https://github.com/deepstreamIO/deepstream.io/pull/612

Thanks for the cleanup ronag!

PR #608, opened 2017-03-25
Prevent Scalar Values in Update Messages

datasage:
This is a better fix for the PR i submitted for the mongo storage layer crash:
https://github.com/deepstreamIO/deepstream.io-storage-mongodb/pull/7

In essence the message specification requires update message to contain an object, but it does not enforce it. Passing a null value through the the set, causes the mongo storage layer to crash. Presumably it would have a similar affect on other storage layers.

To prevent the round trip, this detect should be done within the client itself. I will submit a separate PR for the javascript client.

### DIFF
--- a/src/record/record-transition.js
+++ b/src/record/record-transition.js
@@ -5,6 +5,7 @@ const JsonPath = require('./json-path')
 const RecordRequest = require('./record-request')
 const messageParser = require('../message/message-parser')
 const messageBuilder = require('../message/message-builder')
+const utils = require('../utils/utils')
 
 /**
  * This class manages one or more simultanious updates to the data of a record.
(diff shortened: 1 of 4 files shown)


src/record/record-transition.js
```diff
@@ -147,6 +148,11 @@ RecordTransition.prototype.add = function (socketWrapper, version, message) {
       return
     }
 
+    if(!utils.isOfType(data, 'object') && !utils.isOfType(data, 'array')) {
```
yasserf: We use static references in the code base to types:

https://github.com/deepstreamIO/deepstream.io/blob/f69434acfbaeb21269714a927f3ffc47cdb37625/src/utils/utils.js#L5

Wondering if we should move this into a central place so that we can reference them, or maybe even have a 

`
utils.isOfTypeObject(data)
utils.isOfTypeArray(data)
`

yasserf:
Hey, says you are a first time contributor, would you be okay signing our CLA?

https://deepstream.typeform.com/to/PwwNd5

Thanks!

datasage:
The agreement mentioned a pdf version of the form that needs to be signed, but there was no link to that? Is that still required?

yasserf:
Nope, just signing it via the form should be enough. That was back before the digital age =P

src/record/record-transition.js
datasage: There are several libraries that provide more complete type checking, how far do you want to do take it?

datasage:
I filled out the form.

src/record/record-transition.js
yasserf: We try to keep our dependency tree as small as possible. My point is more that we have OBJECT constants in multiple places and since you added a new convenience method it would be nice to move across to using your utility without having to define strings as arguments.

Not fussed about it though!

yasserf:
Merging. Thank you for raising this!

PR #610, opened 2017-03-25
Permission ordering

yasserf:


src/message/message-queue.js
```diff
@@ -0,0 +1,145 @@
...
+   * @todo The responses from the permissionHandler might arive in any arbitrary order - order them
+   * @todo Handle permission handler timeouts
+   *
+   * @param   {SocketWrapper} socketWrapper
+   * @param   {Object} message parsed message
+   *
+   * @returns {void}
+   */
+  process(rawMessage) {
+    messageParser.parse(rawMessage, this._onMessage)
```
yasserf: @ronag is there a reason this now does a callback? it seems pretty much identical as returning an array of messages.

src/message/message-parser.js
```diff
@@ -48,17 +47,14 @@ module.exports = class MessageParser {
    *                    data: <array of strings>
    *                  }
    */
-  static parse (message) {
-    const parsedMessages = []
+  static parse (message, callback) {
     const rawMessages = message.split(C.MESSAGE_SEPERATOR)
 
     for (let i = 0; i < rawMessages.length; i++) {
       if (rawMessages[i].length > 2) {
```
yasserf: this if statement is confusing me ( theres actually a commit message associated but I can't figure out what it means =( )

src/message/message-parser.js
```diff
@@ -48,17 +47,14 @@ module.exports = class MessageParser {
...
-        parsedMessages.push(this.parseMessage(rawMessages[i]))
+        callback(this.parseMessage(rawMessages[i]), message)
```
yasserf: should we let callback take an error, incase of something like where message is null or not text so deepstream doesn't crash when someone sends a non text frame?

src/message/message-queue.js
ronag: Avoid creating an intermediate array of arbitrary length every time...

src/message/message-parser.js
ronag: No idea... always been like that... https://github.com/deepstreamIO/deepstream.io/blob/master/src/message/message-parser.js#L56

src/message/message-parser.js
ronag: Actually the callback handles null messages already? Non text should be null.

yasserf: Null messages won't work when you do a message.split, why the tests are currently failing

src/message/message-parser.js
yasserf: yup, just wondering why. Oh well

src/message/message-queue.js
yasserf: makes sense!

src/message/message-parser.js
yasserf: Fixed

yasserf:
@AlexBHarley so this is awesome, but we need to add a couple more tests around it before merging it in.

yasserf:
@mrby @AlexBHarley @jdmnd we shouldn't let this PR rot, its really good and enforces a pretty nasty issue when it comes to permissioning. Anyone willing to give it a shot?

yasserf:
alot of it is refactoring which we can ignore because of the clash. The main aspect is the message queue system that feeds messages in the correct order to deepstream

PR #617, opened 2017-03-26
refactor(events): cleanup

ronag:


src/event/event-handler.js
```diff
@@ -1,12 +1,8 @@
-'use strict'
```
yasserf: This is needed for node 4

yasserf:
Can I ask why this PR? =D Other just a refactor ( which is always a plus )

PR #618, opened 2017-03-26
perf(message-parser): Use Set instead of Object.

ronag:


src/message/message-parser.js
```diff
@@ -120,21 +112,20 @@ module.exports = class MessageParser {
...
     const parts = message.split(C.MESSAGE_PART_SEPERATOR)
-    const messageObject = {}
 
     if (parts.length < 2) {
       return null
     }
 
-    if (actions[parts[1]] === undefined) {
+    if (!actions.has(parts[1])) {
       return null
```
ronag: @yasserf: To be honest I don't even think this check is necessary...?

src/message/message-parser.js
```diff
@@ -10,16 +11,7 @@ const C = require('../constants/constants')
...
-  const result = {}
-  let key
-
-  for (key in C.ACTIONS) {
-    result[C.ACTIONS[key]] = key
-  }
-
-  return result
-}())
+const actions = new Set(Object.keys(C.ACTIONS).map(key => C.ACTIONS[key]))
```
yasserf: Is this really a performance improvement? Given actions are static I don't know if using a Set would give any benefits ( or might even make it worse tbh )

ronag: Doesn't matter if actions are static as long as they are dynamically accessed. Let me setup a jsperf.

yasserf: Yeah figured that might the case. Just had this discussion with the core team. A JS perf would be brilliant!

ronag: It's 1-2% faster. However, the code is much nicer :).

https://jsperf.com/object-vs-set-ds/1

ronag: @yasserf: Question remains though, is this check against actions even relevant?

yasserf: Yeah, otherwise invalid actions would be passed around which would result in alot more error checking code everywhere

ronag: @yasserf: Are you sure? As far as I can see there is no harm in it... you get some error logs.

ronag: https://github.com/deepstreamIO/deepstream.io/blob/master/src/rpc/rpc-handler.js#L72
https://github.com/deepstreamIO/deepstream.io/blob/master/src/presence/presence-handler.js#L56
https://github.com/deepstreamIO/deepstream.io/blob/master/src/event/event-handler.js#L52
https://github.com/deepstreamIO/deepstream.io/blob/master/src/record/record-handler.js#L93

ronag: Seems fine 2 me...

yasserf: We are trying to move away from having message parsing edge cases spread across the codebase with a slightly more guarded parser. Plus the fact they can go through permissions with an invalid action is also a bit odd.

In the C++ client we actually ensure that the topic exists, and then that the action is relevant to that topic. It makes the code alot simpler because you only have to worry about invalid context ( like subscribing to same event twice ).

yasserf:
Closing in favor of https://github.com/deepstreamIO/deepstream.io/pull/639

yasserf:
Thanks for the cleanup!

PR #621, opened 2017-03-26
Implemented remoteCount

ronag:


src/cluster/distributed-state-registry.js
```diff
@@ -180,10 +180,15 @@ module.exports = class DistributedStateRegistry extends EventEmitter {
...
       }
-      this.emit('add', name)
     }
 
+    const exists = this._data[name].nodes[serverName]
+
     this._data[name].nodes[serverName] = true
+
+    if (exists) {
+      this.emit('add', name, serverName)
```
yasserf: Can we have a UT for this?

ronag: UT is not my strength. Feel free to take over the PR... I need this for my listener fork but I don't think mainstream DS really needs it at this point... other than actually fulfilling the "API".

yasserf: Will do. Will also take this PR over the fix you made yesterday. Thanks for raising!

PR #622, opened 2017-03-27
Refactor/move apply config

AlexBHarley:


src/record/record-transition.js
```diff
@@ -131,7 +131,8 @@ RecordTransition.prototype.add = function (socketWrapper, version, message) {
     }
 
     try {
-      this._applyConfig(update, message)
+      const config = RecordTransition._getRecordConfig(message)
```
yasserf: static methods can't be private

Issue #624, opened 2017-03-27
Array Operations on Lists


datasage:
We've been starting to test deepstream in production, at least for a few customers. We are using a custom, PHP client, to do updates for a server side. This does work for the most part. But the subscription nature of updating the records does cause issues sometimes.

In this issue I want to discuss lists. In our use case, we have several static lists that we add and remove records from. We never get rid of these lists, we just maintain them.

In situations of concurrency, its possible that the update fails due to another client updating the list record at the same time. Right now our solution is to retry the update, but its inefficient.

So to solve this I am proposing some additions to the message specification that would allow array operations to be directly to update lists without needing to subscribe. In particular for us, being able to do push and removeElement would cover our cases, but i could see adding additional operations like pop and slice.

I would be interested in building this, but I am looking for thoughts as to whether or not others may find this useful.

yasserf:
We are looking to add a concept of version number `-1`. What this will do is bypass the version conflict in record transition, load the last object, apply the patch and then save it with the record version + 1.

This will be useful for adding a HTTP endpoint which will allow post requests to update records without having to subscribe first.

Would that solution work for you? The problem is it doesn't exactly support array operations, and lists are essentially just records under the hood. Theres a whole seperate discussion of us making a LIST api, but that requires alot more work and depends greatly on the underlying tech used ( since we would want to retrieve only the last X records from a list using that API as opposed to records which returns the entire json object, and certain things like REDIS supports that, but others don't ).

datasage:
That would help in another case, but not for lists. I do have a case where the server needs to update the entire object, and it would be easier to not have to subscribe first to do that. For my use case, that is not likely to conflict with a version_exists issue though. Maybe the version of -1 could also bypass any subscription check on the update.

Lists are different for me, in most cases i need to just update the list without reading it as all i need to do is remove or add an element. A list api operations could fix this without significantly changing how lists work.

I can do some research on what Redis, mongo and others support for list operations and propose and API to support it. More than likely we can support everything, it just that some work may have to be done in a connector vs in the storage/message layer itself.

In the meantime, i could add support for the -1 version. That would help simplify my client quite a bit on all other operations.

Using a websocket client in languages like PHP does work. Subscriptions do somewhat complicate things. I think adding more support for more direct operations would allow languages like PHP to work better and could provide the interface needed for a REST interface.

yasserf:
If you could add the `-1` version that would be awesome.

[...]

PR #628, opened 2017-03-28
Added support for -1 version to allow for updates without knowing the version number

datasage:
As discussed in this issue: https://github.com/deepstreamIO/deepstream.io/issues/624

I believe this will work. Not entirely sure how this will interact with the queue system, but it will sort of act as an un-versioned record until its processed. Any records in the queue after that point with version numbers will probably fail on update.


datasage:
In analyzing the problem a bit more.

It may be a good idea to allow for the possibility of upserting an entire record with the update message. This reduces the create/update into a single step.

Alternatively, allow create to be called without the subscription component. Which is probably easier to implement based on current design. That way there is no need to call unsubscribe later.

For me I need to be able to do all operations on records without the needing to subscribe. Extra update packets coming through on the wire make things more complicated. Things work well when i know exactly what packets to expect.

I will explore this a bit more tomorrow, any thoughts on which way you would prefer?

yasserf:
> It may be a good idea to allow for the possibility of upserting an entire record with the update message. This reduces the create/update into a single step.

This makes sense `client.record.set(name, data, callback)` should so an upsert if needed

Alternatively, allow create to be called without the subscription component. Which is probably easier to implement based on current design. That way there is no need to call unsubscribe later.

This makes less sense, since a create is essentially an upsert with an empty object anyways.

```
hasRecord = await client.record.has(name)
if (!hasRecord) {
    client.record.create(name)
}
client.record.set(name, data, callback)
```

> For me I need to be able to do all operations on records without the needing to subscribe. Extra update packets coming through on the wire make things more complicated. Things work well when i know exactly what packets to expect.

I spoke with @ronag about this in a bit of depth. There are interesting edge cases that can occur with setting without subscribing, which allows us to do two options:

Option a:

Send -1 when doing a hard reset (as this PR does and our original thought). This will do exactly what your PR does.

Problems:

Assume two people use the hard set, you'll pass to -1s and both will be written after each other. However if sending a patch then the problem is that will likely result in odd behaviour.

Solution, maybe only support updates via hard set, which limits our API, or..

Option b:

Introduce another message called `HEAD` that returns a version number

example ( totally broken ) code:

```
client.record.set(name) {
  if(!this.recordVersions[name]) {
    this.recordVersions[name] = await getHead(name)
  }
  connection.sendMsg('R|U|{recordName}|{this.recordVersions[name]}|{data}+')
}
```
If a merge conflict is received, the client will then have to reset the version head.

In the case of a normal http call, this will however mean you need to do two. The benefit is we keep our version conflicts working normally, disadvantage is its a bit of an odd approach and since we don't actually get the record content is only there for a *pretty important* sanity check.

I guess the most important aspect of all is ownership. Option B assumes the person doing the sets `owns` the record, and the only reason it actually gets the version number is to verify that. If your 100% sure that nothing else would write to it then using -1 is still a viable solution.

Hope most of that makes sense

ronag:
Here is how CouchDB does it, i.e. option B. 

https://wiki.apache.org/couchdb/HTTP_Document_API#PUT

datasage:
How well locked is the process? If i update a record to version 1 to node one, and send a different version 1 to node 2 at the same time, if the timing is right, could these updates pass checks and then overwrite each other?

datasage:
Starting with the first issue, upsert on update. I don't think an upsert should be assumed. Instead a flag should be passed, like write acknowledgement, as to whether or not an upsert is allowed with that operation. If it is set, and the client has the permission to do so, then a record would be created. I will see about adding support for that over the next couple of days.

The merge issue is an interesting question. In my use case, using -1 would work. We generally establish a lock before allowing the resource to be updated and in the cases where this would be used, the server more or less owns the update.

Due to the lack of support for option 1, we have effectively implemented option 2. Using snapshot to get version number, and then using that to update. Its mostly been working in our testing, although we have seen an odd issue that I haven't been able to track down yet. Having an option to get just the version number would be also a good idea. Having to return the entire record does require a bit of network overhead.

This did get me thinking about how this might behave when you have multiple nodes potentially updating the same record. While it will not affect my use case, since we have a locking mechanism to prevent it. 

It does appear that there might be a race condition when using multiple nodes. If you send the same update to two different nodes at one time, both may retrieve the same version record, and then allow one update to overwrite the other.

datasage:
I added support for a head request. This will give the client some flexibility in determining which option makes the most sense for a given update.

I did see that snapshot does not appear to do a read permission check, is this intended?

datasage:
Never mind my previous message, figured out how permission checks were being done.

datasage:
This should also now support an upsert flag via update. This also checks if the client has permission to create a new record.

yasserf:
> It does appear that there might be a race condition when using multiple nodes. If you send the same update to two different nodes at one time, both may retrieve the same version record, and then allow one update to overwrite the other.

This is correct. There are multiple solutions around this, mainly around using databases that support version control or if you implement versions increments into the cache itself. However we are talking about the time between a read/write to a cache node (tens of thousands a second) so it is very unlikely to happen.

Will look in depth into your PR. Many thanks for raising it!

yasserf:
Looks good!

@AlexBHarley mind giving this another set of eyes before we hit merge?

AlexBHarley:
Pretty happy with how this looks, and I like the use of `upsert` in the config. Just curious @datasage, have you added the `head` call to a client of yours?

datasage:
I do currently do something similar to the head call using snapshot. This has worked fairly well so far, but i do have to create a record if its missing. Which requires a create message followed by an unsubscribe message.

I have not yet updated my client with these changes yet, but doing so will be easy.

yasserf:
Looks good! I believe there will be edge cases around this going forward for clashes and the like... but its a perfect starting point!

ronag:
Um... why do we even need -1 if we have HEAD?

datasage:
In my use case, it would save an extra round trip. The cases where i would use this, it's mostly the server pushing updates to the client, there isn't really a concern about what version the record is.

I see the head option being used where you may have latency in the update, and you want to make sure you update to a certain version.

I still have some issues with updating array records during periods of concurrency. In this case the data does matter. Working out a list operations API would probably solve this for me.

Issue #632, opened 2017-03-29
Delete events do not broadcast with broadcast delay enabled


datasage:
As mentioned here: https://github.com/deepstreamIO/deepstream.io/issues/592

If the broadcast delay is enabled, which is by default. Delete record events do not fire. The subscribers are removed by the point at which they are sent.

My suggestion was to either allow certain broadcasts to override the delay, or pass the current subscribers with delayed broadcast.

datasage:
This is fixed now.

Issue #592, opened 2017-02-20
Multiple Subscribe/Discard bug


valentinvichnal:
Discarding and Subscribing to the same List throws these errors in the client:
```
DS error: lists/todos UNSOLICITED_MESSAGE
DS error: lists/todos MULTIPLE_SUBSCRIPTIONS
DS error: lists/todos ACK_TIMEOUT
DS error: lists/todos RESPONSE_TIMEOUT
```

datasage:
I did run into a similar version of this bug. It would appear that not all subscriptions on a record are removed when the record is removed.

If you recreate the same record and resubscribe to it, you get a multiple_subscriptions error.

I'm attempting to track this down.

ronag:
It's a know bug. There are some async issues with the current implementation.

ronag:
Close in favor of https://github.com/deepstreamIO/deepstream.io-client-js/issues/204

datasage:
I don't think its entirely related to the JS client.

My scenario involves two clients.

1. Client A connects to record.
2. Client B deletes record.
3. Client A everything seems to clean up correctly on client.

Later, without reconnecting client A.

1. Client B recreates record.
2. Client A resubscribes to record.

At this point we get the unsolicited read message and a multiple subscriptions error.

This would seem to indicate that the server kept the subscription for Client A even though the record was deleted.

datasage:
Looking at the wire on "Client A".

I see the delete event being sent when the record is removed. When the record is recreated by Client B, but before i subscribe again with Client A. I'm getting a read update message as if I am still subscribed to the record.

yasserf:
This is interesting. Seems like we should be able to replicate it using your steps in the E2E tests.

> Later, without reconnecting client A.

Does that mean without disconnecting and connecting the client? Do you do that with B?

datasage:
Yeah, the client A stays connected. Client B does disconnect after deletion of the record. and the reconnects to create it again later.

I did dig through the code, nothing stands out for me as to why this is happening. But it does seem like the subscription persists somehow.

yasserf:
Possibly (only for the duration of the client A connection). I'll try and replicate

datasage:
Its possible the change in master fixed it, going to check out the master branch and give a test.

datasage:
[...]

PR #637, opened 2017-04-06
Delete events

datasage:
This adds an option in the sendToSubscribers method to override delay. It should be turned on in any case where subscribers may be removed prior to the broadcast of the event.

I ordered the the noDelay argument as third as the socket argument was considered optional. 

This should resolve the case where delete events do not fire.

Referenced Issues:
https://github.com/deepstreamIO/deepstream.io/issues/632
https://github.com/deepstreamIO/deepstream.io/issues/592
https://github.com/deepstreamIO/deepstream.io/issues/533


yasserf:
Makes perfect sense. Thank you for raising!

PR #644, opened 2017-04-12
Allow record to be read through cache if the cache does not respond.

datasage:
So the reason for this change comes out of a situation I have been running into.

Im using elasticache Redis install on AWS. For some reason, one I haven't yet determined, I lose connection to Redis. Usually, it lasts just a few seconds, but that is long enough to cause some reads to fail.

Now in these cases, the storage layer (a mongo replica set) should be available. This change proposal allows for a read through to the storage layer should the cache layer fail.

This change may not be complete, there could be side effects I'm not aware of. I would like to get some discussion going on this issue to see if we can improve cache handling on some edge cases.

### DIFF
--- a/src/record/record-request.js
+++ b/src/record/record-request.js
@@ -28,7 +28,7 @@ const RecordRequest = function (recordName, options, socketWrapper, onComplete, 
   this._isDestroyed = false
 
   this._cacheRetrievalTimeout = setTimeout(
-    this._sendError.bind(this, C.EVENT.CACHE_RETRIEVAL_TIMEOUT, this._recordName),
+    this._cacheReadThrough.bind(this, C.EVENT.CACHE_RETRIEVAL_TIMEOUT),
     this._options.cacheRetrievalTimeout
   )
 
@@ -36,7 +36,33 @@ const RecordRequest = function (recordName, options, socketWrapper, onComplete, 
 }
 
 /**
- * Callback for responses returned by the cache connector
+ * If cache read times out, allow the read request to read through to the underlying storage mechenism.
+ * 
+ * @param   {String} error  null if no error has occured
+ *
+ * @private
+ * @returns {void}
+ */
+RecordRequest.prototype._cacheReadThrough = function (error) {
+  clearTimeout(this._cacheRetrievalTimeout)
+
+  if (this._isDestroyed === true) {
+    return
+  }
+
+  if (this._options.storageExclusion && this._options.storageExclusion.test(this._recordName)) {
+    this._sendError(error, `timeout exceeded while loading ${this._recordName} from cache:${error.toString()}`)
+  } else {
+    this._storageRetrievalTimeout = setTimeout(
+      this._sendError.bind(this, C.EVENT.STORAGE_RETRIEVAL_TIMEOUT, this._recordName),
+      this._options.storageRetrievalTimeout
+    )
+    this._options.storage.get(this._recordName, this._onStorageResponse.bind(this))
+  }
+}
+
+/**
+ * Callback for responses returned by the cache connector, allows read through if cache lookup fails
  *
  * @param   {String} error  null if no error has occured
  * @param   {Object} record the record data structure, e.g. { _v: 33, _d: { some: 'data' } }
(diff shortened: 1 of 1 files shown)


yasserf:
What happens with writes when your connection drops? Are they cached and flushed on reconnect?

datasage:
I'm not entirely certain what happens in that case.  I'm using the redis deepstream cache layer, and that doesn't seem to anything special in terms of queue up writes in case of a connection failure. I might need to add some more logging to see what is actually happening in those cases.

I do think there is a concern that the cache data does not get invalidated if an update occurs while the cache is down. Then further reads would have stale data, with no mechanism to be eventually consistent. A possible option to set Redis to have a limited cache lifetime might reduce this,

datasage:
Some more thoughts on this. When reading for a record update, we should read directly from storage and bypass cache completely.

Ideally updates should be implemented as a transaction so that we know that the version being written is newer than the version that already exists. 

However, in doing so this would reduce the chance that the cache has stale data and allows a write through that overwrites an existing version. 

If we do get a version_exists error in that case, then there is a probability that the cache is stale. We could also then invalidate the cache data at that point. It may result in some extra storage lookups, but it would reduce the chance that the data is stale.

datasage:
When is a record considered created? It looks like this is currently based on when its written to the cache. Most cache systems generally work with weaker data guarantees when it comes to things like replication (Redis uses asynchronous replication, caches often evict items ).

I've been chasing down some odd cases where records would occasionally get updated with no data. I haven't been able to track that down yet. Logging is lacking in a few places, which i'm working on improving (mostly adding messages at a debug level).

I understand that some uses cases may be ok with a lesser guarantees, but as of right now there are no options for more stronger guarantees.

I would propose that if a storage layer is enabled, that write/change acknowledgments are based on storage layer and not the cache layer. If the cache is shared, an update would invalidate the data stored in the cache and would read through on next read. If the cache is not shared, and invalidate message would be passed through the cluster to invalidate the data for that particular record.

In addition, some sort of optimistic locking should be implemented in the storage layer. This can be done in mongo with validation rules, rdbms's will support it in transactions. This way at least the storage layer can be the source of true as far as the consistency of data.

yasserf:
Hey there, sorry for the long response time for this PR

> When is a record considered created? It looks like this is currently based on when its written to the cache. Most cache systems generally work with weaker data guarantees when it comes to things like replication (Redis uses asynchronous replication, caches often evict items ).

You are correct, it is whenever it is in the cache. The cache eviction is something we are aware of, but users have the option of stating if they want it in the storage layer, and if they do then whenever the data is evicted from cache it will be retrieved again.

> I've been chasing down some odd cases where records would occasionally get updated with no data. I haven't been able to track that down yet. Logging is lacking in a few places, which i'm working on improving (mostly adding messages at a debug level).

We have consciously not added alot of debug messages, because it takes time regardless of whether or not we log at debug ( due to string creation etc ).

What we are toying with is a debug build, which strips out all the debug log lines in production, preferably just using a production flag which the compiler will then optimize and remove as dead code. Just a small heads up incase you see some of that happening.

> I understand that some uses cases may be ok with a lesser guarantees, but as of right now there are no options for more stronger guarantees.

Write acks provide a strong guarantee it has been written to storage. Creating it is another thing obviously, we can now provide guarantees it has been created if using setData instead of getRecord, but that isn't really a solid solution.

> I would propose that if a storage layer is enabled, that write/change acknowledgments are based on storage layer and not the cache layer. If the cache is shared, an update would invalidate the data stored in the cache and would read through on next read. If the cache is not shared, and invalidate message would be passed through the cluster to invalidate the data for that particular record. In addition, some sort of optimistic locking should be implemented in the storage layer. This can be done in mongo with validation rules, rdbms's will support it in transactions. This way at least the storage layer can be the source of true as far as the consistency of data.

Write/change acks are supported in both cache and storage if you use a writeAck flag, the issue is just that if writing to storage fails, other users would still get the data from the cache and be none the wiser, and the client with the failed write ack would need to set the data again forcefully which is an odd action.

I don't know what a decent solution is tbh, we can retry writes to the storage layer potential. However we can't make storage our source of truth because storage is really, really slow. If you compare redis 50k writes (non cluster) to (20M accordingly to AWS) to postgres (20k tops, 100K with biggest aurora machine) you can see how scaling doesn't go hand in hand.

If your use case requires you to have super consistent data you could in theory put your storage layer directly instead of the cache, and then have a cache layer inside of the storage adaptor that allows quicker reads. That way updates will take the hit, but reads should always be super quick. Thoughts?

datasage:
I never did find what was causing the empty records, but have since refactored my approach (essentially splitting the record into two separate ones, one with user writable data, the other with metadata) so it doesn't matter anymore.

On storage vs cache, this largely depends on use case and its also why there are so many different storage engines that emphasize different aspects of the CAP theorem.  When dealing with lots of small data points like stock prices and IOT sensor metrics, loosing a record or having something missed will not matter very much since that data will change within a few seconds anyway. I can see in that scenario, preferring faster writes over consistency. 

In my scenario, i don't deal with a lot of data being active at any given time (1-2 MB per customer), so the speed difference between mongo and Redis is insignificant (Also in my current system configuration, Mongo is local to Deepstream while Reds is not) but accurately storing data, and returning an error properly is.

I think this PR should be closed anyway, the underlying components have changed a bit with the other PR and its not clear yet what the approach should be. I still prefer being able to configure deepstream to ignore cache on records and go directly to storage.

yasserf:
I completely agree. We have a storage exclusion pattern that only allows things to go into cache if its something that is generated. There was a also a proposal that all providers automatically get ignored on the storage layer.

I'm happy you have found a solution for your current scenario, and hopefully we'll find something that fits into the deepstream paradigm that also fixes it at somepoint.

Thanks for the quick feedback!

PR #649, opened 2017-04-14
SocketWrapper Memory Leak

datasage:
As far as I can tell, this array was only being used to provide a connection count. To save performance on adding/removing sockets from the array, I turned this into a counter.

This should also fix a memory leak.

### DIFF
--- a/src/message/connection-endpoint.js
+++ b/src/message/connection-endpoint.js
@@ -369,7 +369,7 @@ module.exports = class ConnectionEndpoint extends events.EventEmitter {
       this.emit('client-connected', socketWrapper)
     }
 
-    this._authenticatedSocketsCounter++;
+    this._authenticatedSocketsCounter++
     this._options.logger.log(C.LOG_LEVEL.INFO, C.EVENT.AUTH_SUCCESSFUL, socketWrapper.user)
   }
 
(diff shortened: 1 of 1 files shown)


ronag:
Nice find!

src/message/connection-endpoint.js
```diff
@@ -369,7 +369,7 @@ module.exports = class ConnectionEndpoint extends events.EventEmitter {
       this.emit('client-connected', socketWrapper)
     }
 
-    this._authenticatedSockets.push(socketWrapper)
+    this._authenticatedSocketsCounter++;
```
ronag: no semi colon

src/message/connection-endpoint.js
```diff
@@ -54,7 +54,7 @@ module.exports = class ConnectionEndpoint extends events.EventEmitter {
     this._ws.on('error', this._onError.bind(this))
     this._ws.on('connection', this._onConnection.bind(this))
 
-    this._authenticatedSockets = []
+    this._authenticatedSocketsCounter = 0
```
ronag: Maybe rename to `_connectionCount`?

yasserf: Given this gets incremented during the authentication callback makes sense to have it as `authenticatedSocketsCounter`

yasserf:
Thanks!

Issue #651, opened 2017-04-16
Record Set JsonPath with numeric parts that are not array indices returned different results on client vs server side


honorcode:
Using localhost (Mac osx) Deepstream.io 2.2.0, Java Client 2.0.8, no plugins

Upon setRecord(recId, "a.b.1", value) the end result of set shows the path created to be: 
{ "a": { "b": { "1": value } } }

Problem: data-sync/getRecord returns this record which does not match and will fail gson.toJson(elem):
{ "a": { "b": [ null, value ] } }

Example Test:
setRecord runtime breakpoint below ->

In UtilJsonPath.java :: setIterateThrough @ line 52
private static JsonElement setIterateThrough(JsonElement element, String path, JsonElement value, boolean delete)
... breakpoint at line 92:            updateValue(value, (JsonElement)parent, token, delete);
where runtime vars: 
    element = {"tickets":{"ticketTexts":{"1":{}}}}
    value = "BDD TESTING"
    token = "1"
    path = "tickets.ticketTexts.1"
    parent = "{"1":{}}"
    delete = false

BUT upon setRecord completing the data-sync'd record and / or subsequent getRecord returns this:
{"tickets":{"ticketTexts":[null,"BDD TESTING"]}}

honorcode:
Btw, much of our data is deeply nested json objs with numerous objects members having numeric fieldnames. These are automated transforms in another ETL application, not possible to refactor easily.

honorcode:
The numeric fieldnames also help significantly as the number of records scales due to memory usage for each fieldname's length for every obj. Smaller fieldnames = smaller objects = smaller data stores = less bandwidth.

honorcode:
We also don't want the added padding of nulls in sparse arrays of slotted values.

redlock:
this is probably related to the issue  #646 I have posted earlier. The fix is also posted. I hope they patch it soon

honorcode:
@WolframHempel FYI the website link for the Contributor License Agreement is broken, I can't do a PR but below is my deepstream.io/master patch with my FIXes for the server-side issues #651 and #646.

honorcode:
honorcode_deepstream_io_patch_fix_json_path_numeric_obj_member_support.diff (ok - .diff file type not supported so I guess I'll include my to .js files (json-path.js and json-pathSpec.js) as .txt files below.

honorcode:
[...]

Issue #646, opened 2017-04-12
Setting An element to an array with index = 0 creates an Object


redlock:
Hi,
While I was trying to set a record array with index 0 like so
```
 record.set(`items[0]`, 51)
```
deepstream create an object rather than an array as follows:
{ items: { "0" : 51 } } 

I traced the error to json-path.js line 40:

```
} else if (this._tokens[i + 1] && !isNaN(this._tokens[i + 1])) {
```
in this conditional that checks for an index, if the value of the index in this._tokens[i + 1] is zero it will fail. 

A quick fix is to replace the conditional with this:

```
if (this._tokens[i + 1] !== undefined && !isNaN(this._tokens[i + 1])) {

```

Hopefully you can patch it up soon. And thanks for your great work with deepstream.

datasage:
Using arrays with systems like deepstream can get messy. Because clients do not have identical states at any given time, making changes to array elements can cause merge issues when two clients update the same array at the same time in different ways.

Lets say for example client a removes element 5, client b updates element 5. If client b gets processed first, the the update will be removed by client a. If client a wins out, client b will actually update was was element 6.

You can only reliably use arrays if you update the entire array in a single operation.

If you want to add remove element, you are best using an identifier that is sequential based on time, but gives each element a unique id that will not be confused in a merge operation.

The client does have a helper method, getUid(), that you can use to generate the identifier.

redlock:
That is true for your scenario. My use case has one publisher that modifies that array and the rest of the clients are subscribers with no ability to modify. Nonetheless, the issue I mention above is still a bug in the json-path.js file as it doesn't behave as intended.

yasserf:
hey @redlock, we will be looking to fix this once easter is over. Thanks for raising!

honorcode:
@WolframHempel @yasserf @AlexBHarley
FYI the website link for the Contributor License Agreement is broken, so I couldn't do a PR but below is my deepstream.io/master patch with my FIXes for the server-side issues #651 and #646.

The two file updates are .txt attachments on #651 (json-path.js/txt and json-pathSpec.js/txt deepstream.io/master 2.2.0)

honorcode:
FYI the website link for the Contributor License Agreement is broken, so I couldn't do a PR but ^^^ is my deepstream.io/master patches with my FIXes for the server-side issues #651 and #646.

honorcode:
@WolframHempel @yasserf @AlexBHarley
[...]

PR #655, opened 2017-04-19
fix(jsonpath): tests and code for #646 and #651

AlexBHarley:
Courtesy of [@honorcode](https://github.com/honorcode), solves #646 and #651

Pretty happy with this, @yasserf do you want to give it a second pair of eyes.

### DIFF
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -33,14 +33,45 @@ JsonPath.prototype.setValue = function (node, value) {
   for (i = 0; i < this._tokens.length - 1; i++) {
     if (node[this._tokens[i]] !== undefined) {
       node = node[this._tokens[i]]
-    } else if (this._tokens[i + 1] && !isNaN(this._tokens[i + 1])) {
-      node = node[this._tokens[i]] = []
+    } else if (this._tokens[i].indexOf('=') > 0) {
+      const arrParts = this._tokens[i].split('=')
+      const token = arrParts[0]
+      const idx = parseInt(arrParts[1], 10)
+
+      if (node[token] !== undefined) {
+        node = node[token]
+      } else {
+        node = node[token] = []
+      }
+
+      if (node[idx] !== undefined) {
+        if (node[idx] instanceof Object || node[idx] instanceof Array) {
+          node = node[idx]
+        }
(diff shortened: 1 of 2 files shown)


src/record/json-path.js
```diff
@@ -51,19 +82,22 @@ JsonPath.prototype.setValue = function (node, value) {
...
  */
 JsonPath.prototype._tokenize = function () {
-  const parts = this._path.split(SPLIT_REG_EXP)
+
+  // makes json path array items a single 'part' value of parts below
+  // 'arrayProp[#]' members transform to 'arrayProp=#' now instead of 'arrayProp.#' previously
+  // see setValue fnc above for special handling of array item parsing vs numeric obj member name
+  // e.g. 'object.1' parsing. this allows for support of parsing and differentiating object
+  // member names that are also numeric values
+  let str = this._path.replace(/\s/g, '')
```
yasserf: this is super expensive, doing three inline regexs and not even compiling them all upfront.

yasserf:
This is a pretty expensive solution.. I would need to look into this more first but I think it can possibly be done in a more performant manner.

yasserf:
I'm guessing this is also a backwards compatability change to how people used the . vs []

This is certainly correctly behaviour, just a bit concerned about impact it may cause merging it in its current state.

AlexBHarley:
So discussed this with @yasserf a bit more. We don't want to do 3 regexes here

```
let str = this._path.replace(/\s/g, '')
str = str.replace(/\[(.*?)\]/g, '=$1')
const parts = str.split(SPLIT_REG_EXP)
```

The first two should be compiled like the third one, but ideally we'd still only want to do one.

A couple of solutions discussed to avoid the `replace(regex, '=$1)`. 

- We can store some metadata somewhere about token when surrounded by `[` and `]`, this way we know it's an array and not a property
- Seemingly better solution: 
  - when we find a token surrounded by `[` and `]`, put it into the token array explicitly as a number (`Number(token)`)
  - otherwise, put it in as a string (`String(token)`)

This way inside the `setValue` function we can see by the type of the token, whether we should be operating on an array or object.

yasserf:
Closing this as its pretty expensive and needs to be addressed in a different manner.

PR #658, opened 2017-04-21
Rule cache fixes

ralphtheninja:
This PR contains:

* a bug fix for the rule cache
* ~~extra `.close()` methods for propagating down to the rule cache when the server "stops"~~
* keep a reference to the timer in `RuleCache` so it can be accessed

### DIFF
--- a/src/permission/config-permission-handler.js
+++ b/src/permission/config-permission-handler.js
@@ -33,9 +33,9 @@ const UNDEFINED = 'undefined'
  *                           to load it from the path provided in options.path.
  */
 const ConfigPermissionHandler = function (options, config) {
-  this._ruleCache = new RuleCache(options)
   this._options = options
   this._permissionOptions = options.permission.options
+  this._ruleCache = new RuleCache(this._permissionOptions)
   this._config = null
   this._recordHandler = null
   this.isReady = false
--- a/src/permission/rule-cache.js
+++ b/src/permission/rule-cache.js
@@ -10,7 +10,7 @@
 const RuleCache = function (options) {
   this._options = options
   this._data = {}
-  setInterval(this._purge.bind(this), options.cacheEvacuationInterval)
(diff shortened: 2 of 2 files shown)


ralphtheninja:
Unit tests pass but not sure why travis complains.

src/permission/rule-cache.js
```diff
@@ -10,7 +10,7 @@
 const RuleCache = function (options) {
   this._options = options
   this._data = {}
-  setInterval(this._purge.bind(this), options.cacheEvacuationInterval)
+  this._interval = setInterval(this._purge.bind(this), options.cacheEvacuationInterval)
```
ralphtheninja: This should not affect anything _but_ the server can at least be closed properly by the caller by doing `clearInterval()` from the outside.

src/permission/config-permission-handler.js
```diff
@@ -33,9 +33,9 @@ const UNDEFINED = 'undefined'
  *                           to load it from the path provided in options.path.
  */
 const ConfigPermissionHandler = function (options, config) {
-  this._ruleCache = new RuleCache(options)
```
ralphtheninja: Passing `options` here is incorrect. `RuleCache` actually wants `options.permission.options`, or `this._permissionOptions`.

yasserf:
Very good catch! That's a little embarrassing. We'll try get the test to pass with clearing the interval to allow the process to die gracefuly

ralphtheninja:
> That's a little embarrassing.

Not easy to catch this. I suspect it's related to changes in configuration, like re-grouping parameters or something like that.

AlexBHarley:
Want to get this into the next release so merging now, we'll get the graceful shutdown in as part of a bigger story later.

ralphtheninja:
FYI. We run an older version of deepstream (before this fix) on one server and the cpu (`docker stats`) is on around 4% when nothing happens. After this fix it goes down to under 0.1%

PR #660, opened 2017-04-21
fix(rpc): fixes #659

### LINKED ISSUE #659: RPC timeouts do not delete RPC from cache
If we have an RPC that times out, it doesn't delete the reference to the RPC which means that when a response is received it throws an exception.

We need to somehow hold a debug reference to RPCs in order so that if it does fail we can hold it for a while longer to log its trail.

ronag:


AlexBHarley:
closing in favour of #662

PR #662, opened 2017-04-24
Fix/#659

AlexBHarley:
@yasserf 

Moved the relevant tests into the rpc handler, because things are now handled one level higher.


src/rpc/rpc-handler.js
```diff
@@ -282,7 +279,9 @@ module.exports = class RpcHandler {
     }
   }
 
-
+  _$onDestroy (correlationId) {
```
yasserf: can we add a lil comment here?

src/rpc/rpc.js
```diff
@@ -55,9 +55,7 @@ module.exports = class Rpc {
   * @returns {void}
   */
   handle (message) {
-        // This guard is for backwards compatability. Having multiple responses or recieving an ack
-        // prior to a response should not actually occur
-    if (this.isComplete === true) {
+    if (this._isComplete) {
```
yasserf: this should never be called, should log it if it gets called or remove it otherwise?

src/rpc/rpc.js
```diff
@@ -149,7 +143,7 @@ module.exports = class Rpc {
   * @returns {void}
   */
   _handleResponse (message) {
-    clearTimeout(this._responseTimeout)
+    this._isComplete = true
```
yasserf: this should be done in destroy no?

PR #674, opened 2017-05-10
Fix MESSAGE_DENIED parameters

jdmnd:
Fix unit tests broken by 0f96c8a


yasserf:
I mean it looks legit! =D

Issue #679, opened 2017-05-14
Unsubscribe Error After Record is Deleted


datasage:
Scenario:

Client A (JS) is subscribed to the a record. Client B (Server, PHP) sends and event to client A which initiates a process that will in part discard the record. After the event is sent, Client B deletes the record.

Client A calls discard, before receiving deleted event from server. By the time discard method gets processed the record does not exist and the server returns an error to the client.

Its possible to work around this, by letting the delete event process instead of calling discard, but the code in my case is used in scenarios were the record is deleted and when its not. So the event would have to differentiate the outcome of the record.

To keep things simpler, it would be better to have the server not send the error at all. Is there a purpose for sending an error on an unsubscribe if the client is already unsubscribed?. It seems to me that due to the asynchronous nature of the system, that we should simply just acknowledge the unsubscribe and just not do anything rather than return an error.

If the message has a purpose with other uses of the subscription registry, I would propose using the silent parameter to squelch the error.

AlexBHarley:
@datasage thanks for raising this. This fix is in version 2.3.0

PR #680, opened 2017-05-14
fix: allow silent parameter to squelch not subscribed error

datasage:
This is in references to issue: https://github.com/deepstreamIO/deepstream.io/issues/679


yasserf:
When will we pass in silent as true?

datasage:
On record unsubscribe. All other cases expect an ackowlegement.

yasserf:
Looks good, thanks!

PR #682, opened 2017-05-22
CREATEANDUPDATE

AlexBHarley:
Allow clients to update a record without being subscribed to it


datasage:
I had added a path for this a couple of months ago by adding an upsert flag to the record metadata.

AlexBHarley:
Yep, one thing we didn't think of though is that this forks the logic quite a bit from what we have currently.

Creating a new action `CU`, in the same vein as `CR` is more like what we currently have. Also removes permissioning from other parts of the server. Ideally permissioning would touch as little as possible.

src/record/record-handler.js
```diff
@@ -241,26 +247,57 @@ RecordHandler.prototype._createOrRead = function (socketWrapper, message) {
   )
 }
 
+RecordHandler.prototype._createAndUpdate = function (socketWrapper, message) {
```
yasserf: missing comment

src/record/record-handler.js
```diff
@@ -241,26 +247,57 @@ RecordHandler.prototype._createOrRead = function (socketWrapper, message) {
...
+  const recordName = message.data[0]
+  const isPatch = message.data.length === 5
+  message.action = isPatch ? C.ACTIONS.PATCH : C.ACTIONS.UPDATE
+  const onComplete = function (record) {
```
yasserf: be interesting to run a simple benchmark to see if this closure being internal actually results in a compiler overhead

yasserf:
@datasage I hope this affect you too badly. We are currently making setting data without a subscription a first class citizen/hot path since we have alot more usecases from the community with publishers than subscribers!

As @AlexBHarley said, because of the way CR works to upsert/read we want to stick to the same concept as much as possible. Moving forward we may even try to move the logic above the permission handler itself rather than the current bypass, but we would need a way to ensure we only do a single read from the cache which we can't easily add.

I hope this doesn't affect you too much, the main difference should be using CU instead of the upsert flag, the code path otherwise is identical.

datasage:
In that case I would prefer that upsert is kept in for one release so i can make the necessary changes to the client and stage the releases. Otherwise I have to do a deepstream upgrade and client upgrade at the exact same time.

From a permission standpoint though, it is a bit strange that one action can require two different permission checks depending on the state of the server. Its convenient, but can be difficult to isolate each system.

I will do a review tonight to see if there is anything else that stands out to me.

yasserf:
@datasage we are considering possibly merging the permissions so that a write by default requires a create. Its all a bit off though, but what we want to move towards in the next few days is the ability to create a hot path that ignores having to read anything from cache or storage and allows the adaptor to decide if it should replace or not on insert (read: using db conflict resolution).

datasage:
I suppose it depends how the permissions are defined. In my opinion, if you have the permission to create, it should imply that you also have the permission to update. The reverse isn't true though, i think you can update without create. 

In this CU message, i think it would be ok to go off of create only.

Being able to go direct to storage would be helpful as long as the cache is invalidated in the flow. Should result in much faster writes.

PR #683, opened 2017-05-24
feature: adding init script

randahlem:


datasage:
There are other init systems than sysvinit. In particular some distributions use systemd instead of sysvinit. Upstart seems to have fallen out of favor.

yasserf:
@randahlem how should we progress with this?

randahlem:
@datasage You are right. The script was initially created for the aws ami which comes with sysvinit. I am currently investigating if you can package both in a useful way.

PR #686, opened 2017-05-26
perf: adding a outgoing buffer for low priority messages

yasserf:


AlexBHarley:
6.10.1 passed and all e2e tests passed on 7 with one failure on a UT.

PR #691, opened 2017-05-28
refactor: turning record request functional

yasserf:


datasage:
Out of curiosity, was is the background on this change?

yasserf:
@datasage we done quite a bit of CPU profiling this release and one of the things we realised was that calling 'new' creates a significant amount of overhead in terms of garbage collection and hence CPU.

Our goal is to keep cpu under 100% at all times, it seems it only peaks only when garbage collection or intensive CPU methods come into play. By removing them slowly this helps alot.

Another massive win was removing .binds, hence passing in the context and the recordName and socketWrapper around rather than binding. This allows the V8 engine to inline the closure functions without deoptimising them due to external context reference which is a huge win (micro optimisation, but we managed to get 10 folds of performance in certain scenarios when everything was added together.

Unfortunately due to the timeout in the record-request its not at its optimal performance, but we are benchmarking a few timeout procedures to see how we can do them in bulk.

Another huge improvement was adding a delayedBroadcast which allows meta events such as errors and acks to be delayed for a short while and send together another request. This has made things like subscriptions much quicker since we only need one FFI on a subscribe instead of two (ack + read).

The biggest bottleneck by far is still the subscription registry. @ronag has made massive improvements but we need to figure out how to improve the checksum in the distributed state registry as well as use less constructs to maintain room state so we can allow thousands of subscriptions to be added in less than a second

datasage:
Thanks for the update.

PR #698, opened 2017-05-31
Osxpkg

randahlem:
2 Steps needed in order to run deepstream after installing:
* mkdir -p /Library/Logs/deepstream
* ~~Start deepstream with`deepstream start -c /usr/local/etc/deepstream/conf/config.yml`~~
* run `deepstream` in the terminal

Note:
Deepstream dir is under /usr/local/etc/deepstream
Plugins reside under /usr/local/etc/deepstream/lib
Logs reside under /Library/Logs/deepstream


scripts/package-conf-osxpkg.yml
```diff
@@ -0,0 +1,140 @@
+# General
```
jdmnd: This config is out of date and should be copied from `conf/config.yml` instead. 
However, this approach of copying the whole file seems like a pain to maintain, (and we've failed to do it for the other packages) so I'd say you should remove all the options that are not the default. Perhaps the packaging file could copy the default config into a neighbour file e.g.
```
cp deepstream/conf/config.yml /opt/.../deepstream/conf/default.yml
```
so that users can see what options can be changed?

yasserf:
@AlexBHarley @jdmnd can we add this path to the default path set?

/usr/local/etc/deepstream/conf/

jdmnd:
@yasserf Why bother hard coding it if we could just search for `./conf/config.yml` relative to the binary?

scripts/daemon/after-install-osxpkg
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+ln -sf /usr/local/build/*/deepstream.io/deepstream /usr/local/bin/deepstream
+mkdir -p /usr/local/etc/deepstream/lib
+cp -r /usr/local/build/*/deepstream.io/* /usr/local/etc/deepstream/
```
jdmnd: Could we make this script create the /Library/logs directory (and thus require sudo)?

randahlem:
Works now @jdmnd. We are good to go basically.

randahlem:
@jdmnd anything left that we need to do?

PR #704, opened 2017-06-02
fix: Fixes inconsistent api usage of state registry

yasserf:
@AlexBHarley


ronag:
What does this fix in practice?

yasserf:
The centralized state registry we used for deepstreamHub via redis. Will hopefully release it as a plugin sooner than later

AlexBHarley:
Looks good. 

But build is failing for what looks like proper reasons.

PR #727, opened 2017-06-16
Feat/multi dimension paths

yasserf:


src/record/json-path.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
-/* eslint-disable no-param-reassign */
+/* eslint-disable */
 const SPLIT_REG_EXP = /[.[\]]/g
```
jdmnd: . doesn't need to be in this expression any more

src/record/json-path.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
-/* eslint-disable no-param-reassign */
+/* eslint-disable */
```
jdmnd: ha

jdmnd:
nice tests dude

PR #733, opened 2017-06-20
handlePermissionStrategy failing on path

Brian-McBride:
When setting up a custom permission handler, the logic was looking for the path at: `permissions.path` and not `permissions.options.path` where it is defined in the sample config.yml file.


yasserf:
Hey brian, thanks for the commit!

It unfortunately is failing the build, would you mind fixing the UTs?

Also, would you be okay with signing [our CLA](https://deepstream.typeform.com/to/PwwNd5)?

Thanks!

PR #750, opened 2017-07-18
feat: enable multiple connection endpoints to be configured at once

jdmnd:
Note: it is not currently possible to configure more than one
subscription registry, so only one connection enpoint can be
configured to allow:
- providing an rpc
- subscribing to an event
- subscribing to record updates
- listening
- subscribing to presence

Configuring more than one plugin with any of these facilities will lead
to undefined behaviour.


AlexBHarley:
+1

yasserf:
LGTM

PR #758, opened 2017-07-24
Feat/listening globs

jdmnd:
Use glob expressions (via minimatch) for listening instead of regular expressions (via RegExp).


src/listen/listener-registry.js
```diff
@@ -842,23 +853,42 @@ module.exports = class ListenerRegistry {
...
   * @returns {RegExp}
   */
-  _validatePattern (socketWrapper, pattern) {
+  _compilePattern (socketWrapper, pattern) {
     if (!pattern) {
       return false
     }
-
-    try {
-      return new RegExp(pattern)
```
ronag: Are you sure there is no overlap between regexp and glob? You might have an expression that is valid both as a glob and as a regexp but with different behaviors.

src/listen/listener-registry.js
```diff
@@ -842,23 +853,42 @@ module.exports = class ListenerRegistry {
...
+      // convert glob pattern to RegExp, returns false if pattern is invalid
+      const regExp = minimatch.makeRe(pattern, {
+        noglobstar: true, // disable ** patterns
+        nocomment: true,  // patterns beginning with # shouldn't be ignored
+        noext: true       // disable "extglob" patterns
+      })
+      if (!regExp && socketWrapper) {
+        this._onMsgDataError(socketWrapper, `Invalid glob expression: ${pattern}`)
+      }
+      return regExp
```
ronag: What's the point of glob if you convert it to a regexp anyway?

Create a new abstraction which can take either a regexp or a minimatch and will call the proper test function.

src/listen/listener-registry.js
```diff
@@ -842,23 +853,42 @@ module.exports = class ListenerRegistry {
...
+      } catch (e) {
+        if (socketWrapper) {
+          this._onMsgDataError(socketWrapper, e.toString())
+        }
+        return false
+      }
+    } else {
+      // convert glob pattern to RegExp, returns false if pattern is invalid
+      const regExp = minimatch.makeRe(pattern, {
+        noglobstar: true, // disable ** patterns
```
ronag: Why?

src/listen/listener-registry.js
jdmnd: there's no ambiguity, if `enableRegexListenPatterns` is set, patterns will be parsed as regular expressions, otherwise they will be parsed as globs

src/constants/constants.js
```diff
@@ -83,6 +83,7 @@ exports.EVENT.INVALID_LEADER_REQUEST = 'INVALID_LEADER_REQUEST'
 exports.EVENT.TIMEOUT = 'TIMEOUT'
 exports.EVENT.LEADING_LISTEN = 'LEADING_LISTEN'
 exports.EVENT.LOCAL_LISTEN = 'LOCAL_LISTEN'
+exports.EVENT.INVALID_PROVIDER_SUBSCRIPTION = 'UNSOLICITED_MSGBUS_MESSAGE'
```
yasserf: invalid enum

src/listen/listener-registry.js
jdmnd: that's the goal, the work here is clearly not optimal and was intended more as an implementation of the api for listening going forwards

jdmnd: we might also be able to move to using https://github.com/micromatch/micromatch which parses more optimally

yasserf:
Closing for now since we need this to match the same pattern matchin as redis

PR #761, opened 2017-07-25
fix: --host and --port options should be used to configure ws server only

jdmnd:
Also adds --httpHost and --httpPort options for the http server


AlexBHarley:
lgtm

PR #780, opened 2017-08-06
subscription registry fixes and improvements

ronag:
So the general idea here is to reduce the number of allocations and lookups and improve iteration performance.


src/utils/subscription-registry.js
```diff
@@ -274,34 +263,35 @@ class SubscriptionRegistry {
...
-
-    if (socket.listeners('close').indexOf(this._onSocketClose) === -1) {
-      socket.once('close', this._onSocketClose)
-    }
+    subscription.sockets.add(socket)
 
     const names = this._names.get(socket) || new Set()
     if (names.size === 0) {
       this._names.set(socket, names)
+      socket.once('close', this._onSocketClose)
```
yasserf: this should be once per socket and not subscription

ronag: @yasserf: It is once per socket.

ronag:
@yasserf: It might be worth it to replace `setTimeout` with a static `setInterval` as to avoid allocating timers on the hot path. What do you think?

ronag:
@yasserf: Any feedback on this? Would be curious to see some performance numbers from your benchmark suite.

yasserf:
Will be looking to get those public soon, generally looks great, thank you!

ronag:
Not sure why the last commit causes failure. Would appreciate some help debugging that.

src/utils/subscription-registry.js
```diff
@@ -398,6 +360,51 @@ class SubscriptionRegistry {
...
+    const subscriptions = this._sockets.get(socket)
+    subscriptions.delete(subscription)
+
+    if (subscriptions.size === 0) {
+      this._sockets.delete(socket)
+    }
+
+    if (this._subscriptionListener) {
+      this._subscriptionListener.onSubscriptionRemoved(
+        subscription.ame,
```
yasserf: typo?

yasserf:
could it be because 

https://github.com/deepstreamIO/deepstream.io/blob/master/src/record/record-handler.js#L658

and u removed the silent flag for unsubscribing?

yasserf:
Sorry for making this take so long, code looks awesome! Hoping to get results on it soon

PR #782, opened 2017-08-09
Fix/resolve permission and auth handlers properly

AlexBHarley:


test/config/config-initialiserSpec.js
```diff
@@ -211,6 +238,33 @@ describe('config-initialiser', () => {
...
+    it('tries to find a custom authentication handler from name', () => {
+      const config = defaultConfig.get()
+
+      config.auth = {
+        name: 'my-custom-perm-handler',
+      }
+
+      expect(() => {
+        configInitialiser.initialise(config)
+      }).toThrowError(/Cannot find module/)
```
yasserf: this error should be custom

PR #786, opened 2017-08-15
Added option for perMessageDeflate to be passed to uWS

daviderenger:


yasserf:
Hey, thanks for raising. What benefit does this provide exactly?

daviderenger:
Hi, adding this flag allows perMessageDeflate to be negotiated with uWS and the messages will be compressed. Currently uWS is not compressing messages right now but it probably will in the future but the messages that are going to uWS will be compressed if they are large enough bu the browser or ws on node. 

I don't know if it is very useful right now and probably not for everyone but adding the option enables the feature to be tested by end users. There might be implementations of deepstream where clients sends really big messages and compressing them would be very beneficial. 

Request: 
....
Sec-WebSocket-Extensions:permessage-deflate; client_max_window_bits
....

Response:
...
Sec-WebSocket-Extensions:permessage-deflate; client_no_context_takeover
...

yasserf:
Cool, makes sense!

You okay signing our CLA?

https://deepstream.typeform.com/to/PwwNd5

yasserf:
Awesome, thanks!

Issue #776, opened 2017-08-02
Record writes through HTTP API don't support arrays at the root level


jdmnd:
The HTTP API refuses record writes where the "data" parameter is an array. 
This makes it hard to interact fully with deepstream lists, which are represented as records with an array at the root level. 
It is also inconsistent with deepstream's `getRecord().set()` and `setData` APIs which allow arrays to be used at the root level.
Record reads over HTTP return an array when the recordName refers to a list, consistent with how other record APIs work.

Related to https://github.com/deepstreamIO/deepstream.io/issues/773#issuecomment-319459165

rbarroetavena:
do you think changing data property type to array + object under 'Record Writes' in the jif schema would be enough to allow writing data parameter as an array? thanks

jdmnd:
@rbarroetavena Yes I believe that should be sufficient. Will try to get this sorted soon, but a PR would be welcome.

rbarroetavena:
PR #788 
thanks

PR #788, opened 2017-08-26
change jif schema to allow writing record data property as an array

rbarroetavena:
related to #776 
it should allow HTTP API record writes where data parameter is an array. useful for list handling.
thanks


jdmnd:
Looks good to me

PR #789, opened 2017-08-28
Listen records in server which you received from clients

Chinafreak:
With this function you're abled to get all records which you receive from clients. This is very important for debug stuffs so you can check if all data of this record look fine.

Of course you can filter those records with pattern/regex which it's already build-in. Feel free to improve it. I just added it because I needed it.

### DIFF
--- a/src/debug/debug-listener.js
+++ b/src/debug/debug-listener.js
@@ -1,0 +1,35 @@
+'use strict'
+
+this._data = {};
+
+/**
+ * Listen record which you receive from clients.
+ * 
+ * @param {String} pattern The pattern to match records which you receive from clients.
+ * @param {Callback} callback callback with parameter recordName for name of record and recordData for data from record.
+ *
+ * @public
+ * @returns {void}
+ */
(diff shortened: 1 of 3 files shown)


Chinafreak:
need a better solution than checking every 1ms at setInterval to receive records data

Chinafreak:
Travis CI make problem. It's not my fault anymore. (Time limit reached)

yasserf:
Hey @Chinafreak 

Many thanks for raising. Unfortunately we won't be able to merge this in, since deepstream is required to run a standalone binary as well as due to to performance issues.

You may be able to create this is a plugin by capturing the data when it enters deepstream via the permission hook btw, would mean you don't need to maintain your own fork if your don't want!

Thanks again!

Chinafreak:
But how about you can enable/disable this settings? If the server runs on standalone, disable this function. If the server runs on node.js or something like that, enable debug modus? Then the performance issues should not be a problem anymore.

Chinafreak:
any update?

PR #790, opened 2017-08-30
Upgrade

yasserf:


src/config/js-yaml-loader.js
```diff
@@ -258,6 +258,6 @@ function getDefaultConfigPath () {
 function replaceEnvironmentVariables (fileContent) {
   const environmentVariable = new RegExp(/\${([^}]+)}/g)
   // eslint-disable-next-line
-  fileContent = fileContent.replace(environmentVariable, (a, b) => process.env[b] || b)
+  fileContent = fileContent.replace(environmentVariable, (a, b) => process.env[b] || '')
```
yasserf: @AlexBHarley isn't this a breaking change?

AlexBHarley: Yes, I was unable to work around this.

However I think having environment variables come in as e.g. `'REDIS_HOST'` instead of `''` is weird behaviour.

PR #815, opened 2017-10-14
Ability for custom permission handler to revoke a granted permission dynamically.

RashadSaleh:
Here, after every permitted operation (record subscription, event subscription, RPC provider subscription), the respective operation handler sends back a function to the permission handler which can be called at any later time to revoke the same permission through emulating an "undo" or doing the reverse operation (record unsubscribe, event unsubscribe, RPC provider unsubscribe), as if it originated, say, from the client.

An example permission handler making use of such a feature is: (***conceptual pseudo code***)
```js
let main_server = new DeepstreamServer({port: 6020, allowRemoteConnections: true});
let helper_server = new DeepstreamServer({port: 6021, allowRemoteConnections: false});

helper_server.start();
helper_server.client = new DeepstreamClient({port: 6021});
await helper_server.login();

main_server.set('permissionHandler', {
  isReady: true,
  canPerformAction: function(username, message, callback, authData) {
   let msg = decode_message(message);
    if (msg.type === 'subscribe to record data') {
      let permission_record_name = 'permission/'+msg.record.type //eg: 'permission/user'
      let rule = await helper_server.client.getRecord(permission_record_name).whenReady;
      let user = await helper_server.client.getRecord('user/'+username).whenReady;
      let result = calculate_permission(rule, user, data); //data is data being read.
      if (result) {
        callback(null, true, (revoke) => { //grant permission;
            rule.subscribe((new_rule) => {
              if (!calculate_permission(new_rule, user, data)) {
                revoke();
                rule.discard();
              }
            });
        }); 
      } else callback(null, false); //reject permission.
    }
  }
}); 

main_server.start();
```

### DIFF
--- a/src/event/event-handler.js
+++ b/src/event/event-handler.js
@@ -33,9 +33,14 @@ module.exports = class EventHandler {
    * @public
    * @returns {void}
    */
-  handle (socket, message) {
+  handle (socket, message, return_revoker) {
     if (message.action === C.ACTIONS.SUBSCRIBE) {
       this._addSubscriber(socket, message)
+      if (return_revoker) {
+        return_revoker(()=>{
+          this._removeSubscriber(socket, message);
+        })
+      }
     } else if (message.action === C.ACTIONS.UNSUBSCRIBE) {
       this._removeSubscriber(socket, message)
     } else if (message.action === C.ACTIONS.EVENT) {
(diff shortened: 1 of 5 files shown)


yasserf:
While this makes alot of sense and is in the right direction we won't be merging it in because we have alot of things coming into v4 and this is quite a big feature. We would also need to find a way it works with valve with is alot more complicated =(

Maybe what you can do is inject a message into deepstream to unsubscribe a certain record. We are looking for hooks to add for these kind of reasons and that maybe a good one.

PR #832, opened 2017-10-26
User already authenticated fix

MuriloFrade:


src/message/message-processor.ts
```diff
@@ -46,7 +46,16 @@ export default class MessageProcessor {
         }, PARSER_ACTIONS.MESSAGE_PARSE_ERROR, message)
         continue
       }
-
+      if (message.topic === TOPIC.AUTH && message.action === AUTH_ACTIONS.REQUEST) {
+        this.services.logger.warn(
+          AUTH_ACTIONS[AUTH_ACTIONS.ALREADY_AUTHENTICATED],
+          'user is already authenticated'
```
yasserf: might be worth mentioning what user here

test/message/message-processorSpec.ts
```diff
@@ -90,4 +90,18 @@ describe('the message processor only forwards valid, authorized messages', () =>
 
     expect(lastAuthenticatedMessage).toBe(message as any)
   })
+
+  it('does not process auth requests', () => {
```
yasserf: this test doesn't make sense, processing a authentication request once should not result in an error message.

yasserf:
@MuriloFrade what is the state of this?

MuriloFrade:
This bug is fixed in the new client/serve release

PR #850, opened 2017-11-21
Use https github url for protocol/binary submodule

clintwood:
This PR updates the original url for the `protocol/binary` git submodule to use github https url.

I'm not totally familiar with the best practices for git submodules but the original url (`git@github.com:deepstreamIO/urp-js`) did not work when I tried update the submodules on my local clone.  The updated url works perfectly using `git submodule update --init protocol/binary`.


yasserf:
Awesome, wanted to do this for a while thanks!

PR #875, opened 2018-01-31
add optional config and config loading via string.

Thomas-P:
There was a problem in the way deepstream used the configuration object. If I want to set some of the parameter, I had also set all the other, because there weren't optional. So I checked if the algorithm worked like 3.1.1. and it themes. That mean, optional parameter won't break things. There was also a problem to set string as the config, like for files. I also fixed this.

### DIFF
--- a/src/typings.ts
+++ b/src/typings.ts
@@ -97,7 +97,7 @@ interface PluginConfig {
   name?: string
   path?: string
   type?: string
-  options: any
+  options?: any
 }
 
 interface DeepstreamPlugin extends NodeJS.EventEmitter {
--- a/src/typings.ts
+++ b/src/typings.ts
@@ -146,41 +146,41 @@ interface LockRegistry {
 }
 
 interface DeepstreamConfig {
-  showLogo: boolean
-  libDir: string | null
-  logLevel: number
-  serverName: string
-  externalUrl: string | null
-  sslKey: string | null
-  sslCert: string | null
-  sslCa: string | null
-  connectionEndpoints: any
+  showLogo?: boolean
+  libDir?: string | null
+  logLevel?: number
+  serverName?: string
+  externalUrl?: string | null
+  sslKey?: string | null
+  sslCert?: string | null
+  sslCa?: string | null
+  connectionEndpoints?: any
 
-  plugins: {
-    cache: PluginConfig
-    storage: PluginConfig
+  plugins?: {
+    cache?: PluginConfig
+    storage?: PluginConfig
   }
 
-  logger: PluginConfig
-  auth: PluginConfig
-  permission: PluginConfig
+  logger?: PluginConfig
+  auth?: PluginConfig
+  permission?: PluginConfig
 
-  storageExclusionPrefixes: Array<string>
-  rpcAckTimeout: number
-  rpcTimeout: number
-  cacheRetrievalTimeout: number
-  storageRetrievalTimeout: number
-  storageHotPathPrefixes: Array<string>
-  dependencyInitialisationTimeout: number
-  stateReconciliationTimeout: number
-  clusterKeepAliveInterval: number
-  clusterActiveCheckInterval: number
-  clusterNodeInactiveTimeout: number
-  listenResponseTimeout: number
-  lockTimeout: number
-  lockRequestTimeout: number
-  broadcastTimeout: number
-  shuffleListenProviders: boolean
+  storageExclusionPrefixes?: Array<string>
+  rpcAckTimeout?: number
+  rpcTimeout?: number
+  cacheRetrievalTimeout?: number
+  storageRetrievalTimeout?: number
+  storageHotPathPrefixes?: Array<string>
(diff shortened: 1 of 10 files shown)


WolframHempel:
Cool, thank so much. Will review and merge shortly


2018-01-31 15:31 GMT+01:00 Thomas Puttkamer <notifications@github.com>:

Thomas-P:
I fixed the resulting TypeScript errors and think the way how the config object work and how it will give to sub modules should be checked. It is better to only give theese values to a sub module, which they need. I fixed it with a workaround by checking the types. But it could have some side effects which are already in the code, because of giving all modules the global config instead of the server name, where it is needed. Hopefully this will be fixed in a refactoring. And I am really sorry about the typo change of one of the files. But the typo, which is used for documentation is a little bit suspect for me and as I copied the correct version from my test environment it broked up.

Thomas-P:
PS: Desctructoring and Default parameter could be also a good way to handle the config object. This is much clearer to read and is only a hint for possible future refactorings.

src/typings.ts
```diff
@@ -146,41 +146,41 @@ interface LockRegistry {
...
-  externalUrl: string | null
-  sslKey: string | null
-  sslCert: string | null
-  sslCa: string | null
-  connectionEndpoints: any
-
-  plugins: {
-    cache: PluginConfig
-    storage: PluginConfig
+  showLogo?: boolean
```
yasserf: The problem is internally we use a deepstreamconfig where non of these are optional, and a user can pass in a config object where they all are because they get merged in from default options.

Because of that I would argue (poorly) that we would need to have an external config object for typescript and an internal one rather than making the codebase have to check if it exists everywhere.

Thomas-P: Hi, I know what you mean, but using it with a type definitions mean that you must provide a full configuration object. Otherwise TypeScript will throw an error. The assertions, that all these properties exists is program logic and should be handled like this. And there are the checks, that these values exists. 

A solution could be to provide two interfaces and a check and transforming function. Typescript will notice in it, that you handle the constraint of existence and then it will accept the return type, where all properties are well set. This could be combined with an [Notification](https://martinfowler.com/eaaDev/Notification.html) pattern, M. Fowler described.

A quick example how a Transformation could work:

```typescript
interface ConfigurationOptional {
  value1?: number
  value2?: string
}

interface ConfigurationForce {
  value1: number
  value2: string
}

const assertType = <T>(defaultValue: T, givenValue?: T): T => {
  return typeof givenValue !== 'undefined' ? givenValue : defaultValue;
}

const defaultObject: ConfigurationForce = {
  value1: 2,
  value2: '3'
}

const assertConfiguration = (a: ConfigurationOptional): ConfigurationForce => {
  return {
    value1: assertType(defaultObject.value1, a.value1),
    value2: assertType(defaultObject.value2, a.value2)
  }
}
```
As you can see, you are able to transform from an optional to a forced constraint object. It is much more readable and you have the checks where you need them. You also have the opportunity to replace the merge method by subsequencing objects. It provide a fully checked object given by design. There have to be some rewrites but not so much. 

I hope it will help a little bit. If there are any questions, feel free to ask.

yasserf: Agreed, which is what I meant for:

> Because of that I would argue (poorly) that we would need to have an external config object for typescript and an internal one rather than making the codebase have to check if it exists everywhere.

The external config option will allow everything to be optional, the internal one will be the one you currently modified (as it is used everywhere).

The assertConfiguration method is already done within the configInitialiser, just need to return a different type.

Thomas-P: Hi,

the actual problem I see is, that you only implicit set values to an object but cannot predict that it happened. This makes it painful to analyze what happened, when something went wrong. This could happen by refactoring something in the modules or forget something to implement and so on. So we are not able to predict, that a function, which we call do what we expect from then todo.
A quick example:
```typescript
// bad
const configLoader = (cfg: any, defaultCfg: any) => {
    doSomething1(cfg, defaultCfg);
    doSomething2(cfg, defaultCfg);
    doSomething3(cfg, defaultCfg);
    doSomething4(cfg, defaultCfg);
    return cfg;
}

// good
const configLoaderGoodWay = (cfg: any, defaultCfg: any) => {
    if (!cfg) {
        return defaultCfg;
    }
    return {
        key1: doSomething1(cfg.key1, defaultCfg),
        key2: doSomething2(cfg.key2, defaultCfg),
        key3: doSomething3(cfg.key3, defaultCfg),
        key4: doSomething4(cfg.key4, defaultCfg),
    }
}
```
You can argue, that both of them do the same, but you cannot verify that there are doing. This is what TypeScript wants to force. If you have an optional value, than create an new object with another type of interface. That helps also to avoid side effects, I which are hard to check. 

And as you see, give your `doSomething` methods only the information they need to work.

If you wish I could make a few quick refactorings, to the problems I see. They will not break the the functionality, I think. So let me know what you are thinking about.

yasserf:
Closing, typescript partials will help solve this for us

PR #912, opened 2018-09-26
Add implementation and test for write ack problem

CatEars:
We noticed that there was a problem when running setDataWithAck from the client and traced the problem here. The main problem is that there are two different callbacks that both use the same message, but one of the callbacks change the message, making it fault in the second one.

First `this.services.storage.set(...)` is called [here](https://github.com/deepstreamIO/deepstream.io/blob/v4/src/record/record-transition.ts#L294), this may be an async operation. This function in turn calls `this.onStorageSetResponse(...)` when it is finished, found [here](https://github.com/deepstreamIO/deepstream.io/blob/v4/src/record/record-transition.ts#L389).

Second, `this.services.cache.set(...)` is called [here](https://github.com/deepstreamIO/deepstream.io/blob/v4/src/record/record-transition.ts#L303), with a callback to `this.onCacheSetResponse(...)` when it is finished, found [here](https://github.com/deepstreamIO/deepstream.io/blob/v4/src/record/record-transition.ts#L364).

If the call to `onStorageSetResponse()` is called after `onCacheSetResponse()` then there is a possibility that `correlationId` and `isWriteAck` will have changed, because of [here](https://github.com/deepstreamIO/deepstream.io/blob/v4/src/record/record-transition.ts#L372), causing `onStorageSetResponse()` to not properly acknowledge the original message.

I solved it with this implementation and added a test for it, emulating an async `storage` and sync `cache` which causes the fault to trigger.

### DIFF
--- a/src/record/record-transition.ts
+++ b/src/record/record-transition.ts
@@ -368,22 +368,16 @@ export default class RecordTransition {
     if (error) {
       this.onFatalError(error)
     } else if (this.isDestroyed === false) {
-      // // Get values that should not be broadcast
-      const { isWriteAck, correlationId } = this.currentStep.message
+      const copiedMessage = { ...this.currentStep.message, isWriteAck: false }
+      delete copiedMessage.correlationId
 
-      // // Delete values that should not be broadcast
-      this.currentStep.message.isWriteAck = false
-      this.currentStep.message.correlationId = undefined
       this.recordHandler.broadcastUpdate(
         this.name,
-        this.currentStep.message,
+        copiedMessage,
         false,
         this.currentStep.sender,
       )
 
-      // Restore the message for other callers to use
-      this.currentStep.message.isWriteAck = isWriteAck
-      this.currentStep.message.correlationId = correlationId
       this.next()
     } else if (this.steps.length === 0 && this.pendingCacheWrites === 0 && this.pendingStorageWrites === 0) {
       this.destroy(null)
--- a/src/record/record-transition.ts
+++ b/src/record/record-transition.ts
@@ -368,8 +368,12 @@ export default class RecordTransition {
     if (error) {
       this.onFatalError(error)
     } else if (this.isDestroyed === false) {
+      // // Get values that should not be broadcast
+      const { isWriteAck, correlationId } = this.currentStep.message
+
+      // // Delete values that should not be broadcast
       this.currentStep.message.isWriteAck = false
-      delete this.currentStep.message.correlationId
+      this.currentStep.message.correlationId = undefined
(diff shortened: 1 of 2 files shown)


src/record/record-transition.ts
```diff
@@ -368,15 +368,22 @@ export default class RecordTransition {
...
       this.onFatalError(error)
     } else if (this.isDestroyed === false) {
+      // // Get values that should not be broadcast
+      const { isWriteAck, correlationId } = this.currentStep.message
+
+      // // Delete values that should not be broadcast
       this.currentStep.message.isWriteAck = false
-      delete this.currentStep.message.correlationId
+      this.currentStep.message.correlationId = undefined
       this.recordHandler.broadcastUpdate(
```
yasserf: We could do

`{ ...mesage, isWriteAck: false, correlationId: undefined }`

I think the cost of assigning a property vs the cost of object creation isn't much different anymore. Thoughts?

yasserf:
Looks like there are some CI issues, will take a look later on today so that PRs are green and ready to go. 

Thank you and @daviderenger for raising, we needed a bit of a push to get v4 out the door!

src/record/record-transition.ts
CatEars: I think it would be wisest to write it in the most clear and concise way possible, and then I think doing what you suggest would be good. Fixing it!

CatEars: I fixed it, but a few tests expected `correlationId` to not exist (`{}` vs. `{correlationId: undefined}`. So in the solution I unpack the original message and specifically delete `correlationId`.

CatEars:
I had some more looking into this and found that a term change for `requestorData` and `requestorName` would be harder than anticipated. Inside of the protocol there are keys defined that basically say what fields in the messages are "allowed" to be sent for "meta". There is also 

In https://github.com/deepstreamIO/urp-js/blob/4b846b3a363977de43cd8bc7c612920318e0eaa2/src/message-builder.ts#L78 we see that only META_KEYS are put into the sent message, otherwise they need to be part of the `data`/`parsedData` field. Obviously we cannot let a user configure what goes into META_KEYS, except the default of `requestorName` and `requestorData`. I do not think that it is a bright idea to start tinkering with how the `data`/`parsedData` looks like, since it can be both null and an object and some RPC_ACTIONS want it to be null.

I am thinking that the option to rename it should be completely removed from the server and if a renaming option should be used then it should be added on the client. I'll update this PR in the afternoon, removing the option to rename `requestorName` and `requestorData`

yasserf:
Thanks!

PR #913, opened 2018-10-05
Make uws terminate dead sockets

CatEars:
We investigated an issue we had with presence and found that uws never terminated the socket, even tough the socket did not answer on several ping broadcasts. The `startAutoPing()` method of uws will automatically remove sockets that have not sent anything since the last ping and was used in Deepstream v3. Taking advantage of this I created a hotfix that terminates dead sockets that have not answered for `2 * this._getOptions('heartbeatInterval')`. I explain it a bit more in the comment included in the pull request and why I used twice the heartbeat interval. The reason I included both the `startAutoPing()` and `broadcast()` is because it seemed the `startAutoPing()` did not send anything to the group (which sounds like a bug in `uws`, but could also be a bug somewhere along the way), that way we force the ping to be sent and with `startAutoPing()` running we will remove any dead sockets that answered the ping from the `broadcast()`

### DIFF
--- a/src/message/uws/connection-endpoint.ts
+++ b/src/message/uws/connection-endpoint.ts
@@ -164,6 +164,31 @@ export default class UWSConnectionEndpoint extends EventEmitter implements Conne
     uws.native.server.group.onPong(this.serverGroup, () => {})
     uws.native.server.group.onConnection(this.serverGroup, this._onConnection.bind(this))
 
+    // When investigating the auto ping we found that it did not (for some
+    // reason) properly send to the group (even tough the code looks like it
+    // should). The result of this is that it finds every client connected to
+    // the group to have disconnected on their end. But since the clients are
+    // actually online (they just did not receive a ping message) they will
+    // reconnect immediately. We run the auto ping, which cleans up the dead
+    // connections at half the rate of the actual heartbeat. Sending them at the
+    // same rate can give the following scenario, bars are timepoint of events:
+
+    //
+    // | First ping is sent
+    //  | pong is received on server
+    //   | Server starts new heartbeat interval. If no pong arrives from client, will terminate it
+    //                      | Second ping is sent
+    //                        | Server terminates socket
+    //      slight network lag | causes pong to arrive to late
+    // ----------- time -------------->
+
+    // For this reason we use 2x heartbeat interval to terminate websockets,
+    // giving them at least 1 * heartbeatInterval to respond, but finding they
+    // are offline might take up to 2x heartbeatInterval
+    uws.native.server.group.startAutoPing(
+      this.serverGroup, 2 * this._getOption('heartbeatInterval'), this.pingMessage
+    )
+
     this.pingInterval = setInterval(() => {
       uws.native.server.group.broadcast(this.serverGroup, this.pingMessage, true)
     }, this._getOption('heartbeatInterval'))


src/message/uws/connection-endpoint.ts
```diff
@@ -164,6 +164,31 @@ export default class UWSConnectionEndpoint extends EventEmitter implements Conne
...
+    //   | Server starts new heartbeat interval. If no pong arrives from client, will terminate it
+    //                      | Second ping is sent
+    //                        | Server terminates socket
+    //      slight network lag | causes pong to arrive to late
+    // ----------- time -------------->
+
+    // For this reason we use 2x heartbeat interval to terminate websockets,
+    // giving them at least 1 * heartbeatInterval to respond, but finding they
+    // are offline might take up to 2x heartbeatInterval
+    uws.native.server.group.startAutoPing(
```
yasserf: does this not conflict with the ping interval? Might want to removed line 192->194?

CatEars: We have tested `startAutoPing` and found that it does not in fact send pings, but as far as I can tell from the source code it should (guessing it is a bug?)

I created a test repo that shows this and what happens instead at https://github.com/CatEars/uws-auto-ping-test

Running `build.sh` in the root repo will build a deepstream client that prints every message and two servers, one with auto ping and `pingInterval` and one with only auto ping.

The output from running without the broadcast looks like this:

client:
```
Got parseResults: [ { topic: 2, action: 3, isAck: false, isError: false } ]
Got parseResults: [ { topic: 3, action: 2, isAck: false, isError: false } ]
logged in!
Got parseResults: [ { topic: 2, action: 3, isAck: false, isError: false } ]
Got parseResults: [ { topic: 3, action: 2, isAck: false, isError: false } ]
Got parseResults: [ { topic: 2, action: 3, isAck: false, isError: false } ]
Got parseResults: [ { topic: 3, action: 2, isAck: false, isError: false } ]
Got parseResults: [ { topic: 2, action: 3, isAck: false, isError: false } ]
Got parseResults: [ { topic: 3, action: 2, isAck: false, isError: false } ]
```

server:
```
...
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
AUTH_SUCCESSFUL | open
CLIENT_DISCONNECTED | open
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
AUTH_SUCCESSFUL | open
CLIENT_DISCONNECTED | open
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
AUTH_SUCCESSFUL | open
CLIENT_DISCONNECTED | open
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
AUTH_SUCCESSFUL | open
CLIENT_DISCONNECTED | open
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
```
So that is several logins and log outs (because uws thinks the socket is dead since it has not responded to auto pings, which uws fails to send, for some unkown reason)

The output with the broadcast activated:

client:

```
Got parseResults: [ { topic: 2, action: 3, isAck: false, isError: false } ]
Got parseResults: [ { topic: 3, action: 2, isAck: false, isError: false } ]
logged in!
Got parseResults: [ { topic: 2, action: 1, isAck: false, isError: false } ]
Got parseResults: [ { topic: 2, action: 1, isAck: false, isError: false } ]
Got parseResults: [ { topic: 2, action: 1, isAck: false, isError: false } ]
Got parseResults: [ { topic: 2, action: 1, isAck: false, isError: false } ]
```

server:
```
...
INCOMING_CONNECTION | from undefined (127.0.0.1)
REQUEST | 127.0.0.1
AUTH_SUCCESSFUL | open
```

Issue #924, opened 2018-12-07
Send random bytes crash server


phmasters:
When send  random bytes trough RPC call, it crashes the server.
the problem resides in message-processor.js

the line 54 must be placed after line 65 to prevent the problem. I'll do a pull request, but dont have the time now.

Thx

rzhade3:
Hi @phmasters, have you been working on this? If not, I can submit a PR to fix this issue.

yasserf:
Thanks for the fix 😁

PR #927, opened 2018-12-27
Moving null checker before reference

rzhade3:
Currently `parsedMessage` is being referenced before it is being checked for NULL. This can lead to errors, as mentioned here: https://github.com/deepstreamIO/deepstream.io/issues/924.

This PR fixes the issue by moving the check for null before the reference, in order to log any errors and fail gracefully.


indeyets:
@yasserf can you check this please?

yasserf:
👍 Thank you for the PR!

Apologies for the delay, things have been a little hectic over New Years.

PR #949, opened 2019-04-28
Fix for iterating over headers

remaininlight:
I was encountering a bug in v4.0.0-rc.6 where desiredHeaders passed to getHeaders in uws/connection-endpoint.ts can be undefined, causing a crash when it was attempted to iterate over it. This fix adds a function in utilities to test for null and iterability before proceeding with iteration.


yasserf:
Thanks for raising. The issue with this is the fact it ain’t set in the default options however, not that it can’t be null! There’s a fix open currently in a separate branch which needs to be put in.

Thank you though, will be sure to get a fixed release out ASAP!

PR #950, opened 2019-05-11
[fix] set header on options request when all origins are allowed

jaime-ez:
Hi! When upgrading to the latest release started getting CORS errors. When the client makes an option request, if allowAllOrigins is true, the 'Access-Control-Allow-Origin' header was not being set.

### DIFF
--- a/src/message/http/server.js
+++ b/src/message/http/server.js
@@ -277,7 +277,9 @@ module.exports = class Server extends EventEmitter {
         return
       }
     }
-
+    if (this._config.allowAllOrigins) {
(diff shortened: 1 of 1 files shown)


yasserf:
Any chance you want to raise against V4 branch too? 🙈

jaime-ez:
On v4 there's no problem. The previous node http server implementation was setting the headers right. Keep in mind when updating to uws server.

yasserf:
Ah sorry I forgot I didn't swap that to the new uws server yet!

PR #964, opened 2019-06-30
fix ssl params

lyollix:
server = new uWS.SSLApp(...) want only file names


yasserf:
hey @lyollix, thanks for the PR. There's actually a bit more work involved (breaking tests + lack of standard) so it makes more sense to move the logic into each endpoint.

For now, would you be okay running it behind a ssl termination endpoint like nginx instead?

yasserf:
Closing this as theres alot more work involved, thanks for raising though, will have it out in a future release!

PR #979, opened 2019-08-18
generate declaration (.d.ts) files during build

Vortex375:
Added
`"declaration": true`
to tsconfig.json so that .d.ts files are generated during build.

I had to work around a compile error that occurs only when building with declarations enabled. All of the files in bin/ would fail with
```
TS4023: Exported variable '...' has or is using name 'local.Command'
from external module "node_modules/commander/typings/index" but cannot be named.
```

The only work-around I could find was to introduce an empty interface that extends the "Command" type.


yasserf:
Awesome thanks!

PR #1012, opened 2019-10-07
Update js-yaml-loader.ts

V0lk3n:
Modify "they will be defaulted to default values" to "they will be reverted to default values"


V0lk3n:
I was exausted and want to said "reseted" but i put it beacause "defaulted to default value" is too repetitive ^^ but do it as you want :)

yasserf:
ah cool okay, in that case it would probably be best to say:

```
If some properties are not set they will be defaulted to the values defined in the defaultOptions.js file
```

V0lk3n:
Yes perfect! :)

PR #1016, opened 2019-10-10
Update README.md

V0lk3n:
The two dev link at line 5 redirect to error 502, remove both link or update them


yasserf:
Thanks, but thats just a temporary issue that effects everyone using that badge 😅

PR #1063, opened 2020-04-23
load permissions from file when config is not yaml file

jaime-ez:
Valve permissions where not being loaded when server started from script passing options `new DeepstreamServer(options)` neither when passing a `.js` config file in cli.


yasserf:
Sorry that close was accidental!

yasserf:
I'm afraid this is failing the unit tests

jaime-ez:
fixed

yasserf:
releasing as 5.0.15

PR #1080, opened 2020-08-09
[fix] use upgrade property for uws http server

### LINKED ISSUE #1079: uws above version v18  causes uws http server to crash
The parameter `request` is no longer available since this commit

https://github.com/uNetworking/uWebSockets.js/commit/b4df9ebf5f4c3ecc884d5f5e0df1cd04b8caa7c3#diff-01c1a4e28835a7d097ce62009d6e3db1L97-R127

jaime-ez:
fixes #1079 

@yasserf e2e:v3 tests are not passing, should they be removed from the husky script? Since they're alredy removed from CI...of course it's not ideal but given the scarce ressources...if you have any hint on how to fix them let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Use upgrade property for uWebSockets http server.
 
+### Misc
+
+- Do not run e2e:v3 test on the pre-push hook
+
 ## [5.1.2] - 2020.07.05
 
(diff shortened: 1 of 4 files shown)


yasserf:
Yeah let’s just ignore them. If someone uses a V3 combo and has issues they can raise a GitHub issue.

Issue #1041, opened 2020-01-21
deepstream@5.0.5 File Authentication


caracal7:
**deepstream.io server**
```
const fs = require('fs');
const path = require('path');
const { Deepstream } = require( '@deepstream/server' );

const DeepstreamServer = new Deepstream({
	logLevel: 'DEBUG',
	auth: [{
		type: 'file',
		options: {	
		       users: __dirname +`/users.yml`,   //.  users: fs.readFileSync(__dirname +`/users.yml`, 'utf8'),
    		       hash: 'md5',				
    		       iterations: 1,				    
    		       keyLength: 32,				
		}
	}]
});

```
**users.yml**
```
server:
  password: 202cb962ac59075b964b07152d234b70
  clientData:
    favorite color: blue
  serverData:
    department: admin

chris:
  password: tsA+ks76hGGSGHF8**/JHGusy78=75KQ2Mzm
  clientData:
    favorite color: blue
  serverData:
    department: admin
```
**Node.js client**
```
const { DeepstreamClient }  = require( '@deepstream/client' );
const client = new DeepstreamClient( 'localhost:6020' );
client.login( { username: 'server', password: '123' }, (success, clientData) => {
	console.log(success)
});

```
**deepstream.io server console**
```
undefined
PLUGIN_INITIALIZATION_ERROR | missing password for 0
PLUGIN_INITIALIZATION_ERROR | missing password for 1
PLUGIN_INITIALIZATION_ERROR | missing password for 2
PLUGIN_INITIALIZATION_ERROR | missing password for 3
PLUGIN_INITIALIZATION_ERROR | missing password for 4
PLUGIN_INITIALIZATION_ERROR | missing password for 5
PLUGIN_INITIALIZATION_ERROR | missing password for 6
PLUGIN_INITIALIZATION_ERROR | missing password for 7
PLUGIN_INITIALIZATION_ERROR | missing password for 8
PLUGIN_INITIALIZATION_ERROR | missing password for 9
PLUGIN_INITIALIZATION_ERROR | missing password for 10
PLUGIN_INITIALIZATION_ERROR | missing password for 11
PLUGIN_INITIALIZATION_ERROR | missing password for 12
PLUGIN_INITIALIZATION_ERROR | missing password for 13
PLUGIN_INITIALIZATION_ERROR | missing password for 14
PLUGIN_INITIALIZATION_ERROR | missing password for 15
PLUGIN_INITIALIZATION_ERROR | missing password for 16
[...]

PR #1085, opened 2020-10-13
enable file auth path and multiple authentication stategies

jaime-ez:
#1041 

Enable CombineAuthentication, a very useful feature that was unused and undocumented (to be done).


jaime-ez:
Regarding the multiple auth strategies. Taking a second look to the code base, when a login is unsuccesfull, some services return null or {isValid: false}. If the service returns null, the combine will continue iterating, otherwise it will return. For file and for storage auth strategies there is reportInvalidParameters option that will force {isValid:false}. It is http webhook the one that is breaking this logic. A reportInvalidParameters option should be added to fail or continue for the http webhook and thus keep the combine auth as it was.

Comments?

yasserf:
This looks good to me! Thanks!

jaime-ez:
Ok I'll update the pull soon to reflect the previous proposal

conf/config.yml
```diff
@@ -270,6 +272,8 @@ auth:
   #     iterations: 100
   #     # the length of the resulting key
   #     keyLength: 32
+  #     # fail authentication process if invalid login parameters are used
+  #     # reportInvalidParameters: true
```
yasserf, line 276: looking at the code here, it would make sense the config is passed into the auth strategies via the combiner

i can't see a situation where we would want the reportInvalidParameters to be true in a combined and false normally.

jaime-ez: So if there is more than one auth the combiner should pass the reportInvalidParameters: false to the isValid function ?

yasserf: yup

PR #1086, opened 2020-10-13
Fixed crash caused by TypeError due to missing builder.

redlock:


src/connection-endpoint/websocket/text/text-protocol/message-builder.ts
```diff
@@ -213,8 +213,10 @@ const BUILDERS = {
  */
 export const getMessage = (message: Message, isAck: boolean = false): string => {
   if (!BUILDERS[message.topic] || !BUILDERS[message.topic][message.action]) {
-    console.log(message, isAck)
+    console.trace('missing builder for', message, isAck)
+    return ''
   }
+  
   const builder = BUILDERS[message.topic][message.action]
   if (!builder) {
```
jaime-ez, line 221: Could this check be removed since we are doing it and returning in the previous block?

redlock: Done.

jaime-ez:
looks good to me, could you please bump the version number in package.json, package-lock.json and update the changelog in order to make all relevant changes in one pull ?

Thanks

PR #1134, opened 2023-09-09
Multi arch Docker image builds

daanh432:
This pull requests switches to parallel building of the regular and alpine image building. Both of which will be built for both AMD64 (same as current) and ARM64 (new addition).  [See on Docker Hub](https://hub.docker.com/r/daanh432/deepstream.io/tags)

Using EKS or AKS with ARM based CPUs like Graviton these multi arch Docker images are wonderful to use.

All pipelines were also using node 18 except for the Dockerfile. I think it would be better to use the same node version for all artifacts.

Any existing Docker setup will not break. Everything is still mapped to ``/usr/local/deepstream``.

Can be validated using the test image I have published on Docker Hub.

```sh
docker create -t -p 6020:6020 -p 6021:6021 \
  --name deepstream.io \
  -v $(pwd)/conf:/usr/local/deepstream/conf \
  -v $(pwd)/var:/usr/local/deepstream/var \
  daanh432/deepstream.io:7.0.8
  ```


daanh432:
Considering this project is unmaintained the pull request of this can be considered as irrelevant.

jaime-ez:
Hi! It's not unmaintained, I'm just a slow maintainer!

I'm going to reopen the pull and I'll try to look into it soon, thanks

hugojosefson:
If I understand correctly, this can now be considered merged too! :tada:
@daanh432's docker commit was included in my (now merged) PR #1135, because that was created on top of this PR.

PR #1135, opened 2023-10-15
Upgrade dependencies

hugojosefson:
This builds upon https://github.com/deepstreamIO/deepstream.io/pull/1134 by @daanh432, which uses Node.js 18. 
In this PR, I upgrade (almost) all other dependencies, for security, and for trying to stay up-to-date.

For some dependencies, such as `typescript` v4 → v5, the amount of things to fix after such an upgrade was too great for me at the moment, so I stayed on the latest minor version I could successfully upgrade to.

Some minor, and some major, version upgrades to dependencies required small updates in code, which I included in each such upgrade commit together with the dependency version bump.

I successfully ran this command, at least after every major version upgrade commit, and at the end:

```bash
npm install \
&& npm run lint \
&& bash -euo pipefail ./scripts/tsc.sh \
&& npm test \
&& npm run e2e -- --fail-fast \
&& npm run e2e:uws -- --fail-fast \
&& bash scripts/package.sh true true \
&& node scripts/node-test.js \
&& node scripts/executable-test.js
```

That way, I feel confident this should work.

I have also tested the major upgrade of `commander` manually with the built executable `./build/deepstream`, to make sure it sees the `start` command as the default and chooses the correct command when given another one.

I updated one test to make it work correctly in both Node.js 18 and Node.js 20, because the error message it checks for, becomes longer and more specific in Node.js 20.

---

_Build is green on my fork's GitHub Actions: https://github.com/hugojosefson/deepstream.io/actions/runs/6591970600_


src/config/config-validator.ts
```diff
@@ -291,7 +291,7 @@ export const validate = function (config: Object): void {
   const valid = validator(config)
 
   if (!valid) {
-    const output = (betterAjvErrors(schema, config, validator.errors, { format: 'js' }) as never as betterAjvErrors.IOutputError[])
+    const output = betterAjvErrors(schema, config, validator.errors ?? [], { format: 'js' })
```
hugojosefson, line 294: The new version returns the expected type, thanks to `{ format: 'js' }` in the options, so no cast is needed.

src/config/js-yaml-loader.ts
```diff
@@ -104,7 +104,7 @@ function parseFile<ConfigType = DeepstreamConfig> (filePath: string, fileContent
   const extension = path.extname(filePath)
 
   if (extension === '.yml' || extension === '.yaml') {
-    return yaml.safeLoad(replaceEnvironmentVariables(fileContent)) as unknown as ConfigType
+    return yaml.load(replaceEnvironmentVariables(fileContent)) as unknown as ConfigType
```
hugojosefson, line 107: Functions are renamed, all "safe" now: https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md

src/services/authentication/file/file-based-authentication.ts
```diff
@@ -102,7 +102,7 @@ export class FileBasedAuthentication extends DeepstreamPlugin implements Deepstr
         validateHashingAlgorithm(settings.hash)
       }
     } catch (e) {
-      this.services.logger.fatal(EVENT.PLUGIN_INITIALIZATION_ERROR, 'Validating settings failed for file auth', e.message)
+      this.services.logger.fatal(EVENT.PLUGIN_INITIALIZATION_ERROR, 'Validating settings failed for file auth', { error: e })
```
hugojosefson, line 105: The type error I got here, was that `e.message`, wasn't a valid `MetaData` that the function expected. So the code probably had some (inconsequential?) bug here.

However, the `MetaData` interface allows non-declared keys with `any` value, so I chose to put `e` as the value of a new key `error` here. I hope this is in the spirit of the code!

jaime-ez:
Awesome, I'll check the change and leave some comments

src/services/authentication/file/file-based-authentication.ts
jaime-ez: Ok with extending metadata.
Can you double check if we are actually logging metadata on fatal errors? A quick look at the code makes me think we are not. at least in the standard logger

hugojosefson: Thank you for taking the time to review and comment. I will look into this! 👍

hugojosefson: If you mean if we add metadata to the individual fatal log statements, it seems that only happens in this instance in `file-based-authentication.ts` above:

image

For how any included metadata is handled, I found this, which indicates that metadata apparently is included and sent along in some default logging code: https://github.com/deepstreamIO/deepstream.io/blob/50b2b914ff97b7c913114dc1a611422922dafa42/src/services/logger/pino/pino-logger.ts#L83-L94

Does this answer your question, or were you thinking of something different?

hugojosefson:
Thank you @jaime-ez for taking the time to review and merge this!

jaime-ez:
Thanks @daanh432 and @hugojosefson for your PRs!
Awesome to see some actual work on the project from others.
new release will be available soon, please test and inform any issues.

Cheers

src/services/authentication/file/file-based-authentication.ts
jaime-ez: thats what I meant, the standard logger does not log metadata on fatal error. created issue #1136 as reminder

